PR #100, opened 2015-02-27
TMI-98: PSDMetadata member visibility fix.

JavadocMD:
Based on your sample code, it looks like PSDMetadata.NATIVE_METADATA_FORMAT_NAME should be public.


haraldk:
Great stuff, thanks! 

I only tested from the same package so I didn't notice, sorry.

-- 
Harald K

PR #188, opened 2015-11-11
Fixed an issue with long runlengths in CCITTFax writing

Schmidor:
When a runLength had more than 2560 pixels the remaining runLength part was miscalculated.


haraldk:
Good stuff! 

I'll merge this!

 However, do you have a test case that will fail without this patch, and pass with it? Just so we know we don't re-introduce this bug at a later stage? :-)

-- 
Harald K

Schmidor:
I have a file that provokes this but I'm not comfortable uploading it. I have to see if i find another file or can produce one.

haraldk:
I don't think need a file to test the writer. What about just creating a minimal image in-memory, that has the needed characteristics and try to write it?

-- 
Harald K

PR #200, opened 2015-12-22
add TIFFImageMetadata.getTIFFField

Schmidor:
Adding a method for simper access to TIFFFields than processing the node tree. JAI has a similar method in TIFFImageMetadata.


haraldk:
Looks reasonable to me. :-)

-- 
Harald K

PR #207, opened 2016-01-17
Enable sequence writing on TIFFImageWriter

Schmidor:


Schmidor:
I've missed some pointer updates on compression != uncompressed.

PR #215, opened 2016-02-01
Supporting JPEGLossless

HermannKroll:
Hi,

finally I added the JPEGLossless support to the TwelveMonkeys. My Code wraps the following library to convert the decoded data automatically into BufferdImages.

JPEG Lossless files which are supported by https://github.com/rii-mango/JPEGLosslessDecoder can be read.

Careful: currently only supports
16, 8-bit grayscale and 24 bit rgb conversion for BufferedImages

Best regards,
Hermann Kroll

### DIFF
--- a/imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGImageReader.java
+++ b/imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGImageReader.java
@@ -39,6 +39,8 @@ import com.twelvemonkeys.imageio.metadata.exif.TIFF;
 import com.twelvemonkeys.imageio.metadata.jpeg.JPEG;
 import com.twelvemonkeys.imageio.metadata.jpeg.JPEGSegment;
 import com.twelvemonkeys.imageio.metadata.jpeg.JPEGSegmentUtil;
+import com.twelvemonkeys.imageio.plugins.jpeg.lossless.DataStream;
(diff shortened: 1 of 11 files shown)


haraldk:
Sorry, 

I'm a little swamped up in other work these days, but I will integrate this some time soon and before the next release. 


Thanks,

-- 
Harald K

HermannKroll:
Hi,

no stress. We have our version of the library working.

P.S.
I added me to the pom.xml file.

xiexed:
Hello. I am trying to use your patched version for reading DICOMs, but I've found that it is impossible to get raw decoded data (which is sometimes needed in medical image processing) because the method

    @Override
    public Raster readRaster(int imageIndex, ImageReadParam param) throws IOException {
        return delegate.readRaster(imageIndex, param);
    }

of [JPEGImageReader.java#L929](https://github.com/haraldk/TwelveMonkeys/blob/master/imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGImageReader.java#L929) always delegates to system reader so is not able to read Lossless JPEG.

Maybe there is an easy way to add this functionality? Of course it is the behaviour of TwelveMonkeys `JPEGImageReader` not brought by this patch, but... it is mainly actual in case of this patch.

haraldk:
@xiexed 

Good catch! That seems to be an oversight. I think we need to parse the "headers" and test for lossless encoding before either delegating to the system reader or doing a lossless decoding of the raster only.

Can you open a normal issue for this (even though the change is not integrated yet, it will be), and reference this pull request (#215) in the issue?


Thanks,

-- 
Harald K

haraldk:
Manually integrated with changes.

PR #217, opened 2016-02-01
Catch EOF at reading IFD pointer

Schmidor:
More tolerant parsing for TIFF Writers, which write the metadata at file end and don't care for adding the finalizing IFD pointer to 0.


haraldk:
Looks like a good thing!

I think the test should make sure that the IFD still contains the expected values though, currently the test would pass even if the `EXIFReader` would just silently ignore the entire IFD (or even the entire file).

Other than that, I'll merge this ASAP.

-- 
Harald K

PR #223, opened 2016-02-16
Continue reading on invalid tags

Schmidor:
Read metadata fully even if an invalid tag is found.


eschreiner:
Hi Harald,

do you already have a release date for this fix in mind (e.g. version 3.2.2). I would need this fix for a project I am working on...

Thank you so much & best regards,
Erich

haraldk:
Hi Erich,

My plan is to release 3.3 as soon as possible, the version will include this fix. I don't currently plan on releasing any bugfix releases before that, as my time budget is very limited. For the same reason, I can't make any promises on an exact date.

If you need a fix now, I suggest you download/clone the project and build a pre-release version for use until the official release is out. It should be as easy as:

    $ git clone git@github.com:haraldk/TwelveMonkeys.git
    $ mvn install


Best regards,   

-- 
Harald K

PR #224, opened 2016-02-16
Fix reading srcRegion for OLD_JPEG

Schmidor:


haraldk:
Looks good!

Thanks! 
I'll try to merge your pull requests tomorrow or so. :-)

(PS: I suspect there's a similar issue for "new" JPEG as well, except the logic is more complicated due to striping/tiling, if you like to have a stab at it.)

-- 
Harald K

Schmidor:
Could be, but at least it doesn't manifest itself as an error or I suspect I would have spotted it :-)
If there is an issue in "new" JPEG this could also happen when manually reading the JPEG Tables on OLD_JPEG, as tile reading looks similar in both.

haraldk:
True (and now I feel bad for the code duplication as well..). ;-)

-- 
Harald K

PR #225, opened 2016-02-17
More tolerant CCITT reading

Schmidor:
More tolerant CCITT reading if the stream contains less lines than are tried to read


haraldk:
Hi, realized your fix resulted in black lines, rather than white at the end, so I changed that (as the result seems more in line with other tools).

Also added a test case to the `CCITTFaxDecoderStreamTest` to verify correct image data.

Otherwise, looks good!

-- 
Harald K

PR #227, opened 2016-02-19
Support OLD_JPEG and SubIFDs in TIFFUtilities

Schmidor:
Support OLD_JPEG and SubIFDs in TIFFUtilities on merging and splitting pages.

Is it ok to switch the default IFD from the reader to be editable?


haraldk:
Well... If we really want IFD to be mutable, that will require some more work to keep it sorted, avoid duplicates etc. So I'd rather avoid making it mutable. 

As the `EXIFWriter` can already write any `Collection<Entry>`, what do you really gain by it?

-- 
Harald K

Schmidor:
It would simplyfie the usage to just replacing entries instead of first copying the entries to a new collection.
I also really would like to have a "TIFFImageMetadata.setTIFFField" complementing the "TIFFImageMetadata.getTIFFField". But if this requires further steps, I would change it back to filling the collection here and make myself an Ticket to revisit this later.

Issue #229, opened 2016-02-23
TIFFImageWriter writes uncompressed files with incorrect "StripByteCounts" field


haraldk:
The TIFFImageWriter does not correctly calculate `StripByteCounts` for uncompressed TIFFs (the default) if the bits per sample is not 8. Most tools will just ignore the value (as in the "Warning: Bogus StripByteCounts field") and compute it correctly anyway if the data is uncompressed, but it does cause problems for certain tools on the Windows platform.

See #226 for background & discussion.

THausherr:
Why is "uncompressed" the default? Why not set G4 for bitonal and LZW for color images? These would be the best possible non-lossy formats for "dumb users".

haraldk:
A lot of similar thoughts today... I just commented on #241 that we should consider changing these defaults. :-)

Will you file a separate issue for changing the default compression?

The reason for uncompressed is that it's the fastest, it's baseline (most compatible) and just works. The problem with defaulting to CCITT/G4 and LZW is they are not baseline options, even though I'm sure 99% of all TIFF software supports that.

In any case, a goal of the TIFF plugin is to be a drop-in replacement for the JAI TIFF plugin, so the defaults should be the same. But I'm sure we can tweak this by a system property or similar mechanism.

-- 
Harald K

THausherr:
No, I am not planning to create an issue now, as we are not yet using the TIFF writing from twelvemonkeys (but we might in the future). We're still using JAI with some options.
https://svn.apache.org/viewvc/pdfbox/trunk/tools/src/main/java/org/apache/pdfbox/tools/imageio/TIFFUtil.java?revision=1709449&view=markup&sortby=date

haraldk:
As far as I could tell from looking at the JAI source code, the defaults are always uncompressed.

-- 
Harald K

haraldk:
Fixed in 04a3915.

-- 
Harald K

Issue #228, opened 2016-02-23
TIFFImageWriter does not correctly write images with "sampleModelTranslate"


haraldk:
TIFFImageWriter does not correctly write images with "sampleModelTranslate" (ie. as returned by `BufferedImage.getSubimage(...)`). 

The writer is accessing the pixel data directly without taking the offset and the original pixel stride into account, and as a result, the written image looked skewed (or in other words, just completely messed up).

See #226 for background & discussion.

haraldk:
Failing test case for the `TIFFImageWriterTest`:

````java

    @Test
    public void testWriteCropped() throws IOException, URISyntaxException {
        // TODO: Test 2, 4, 8 bit palette data + gray + ushort + float types?
        List<URL> testData = Arrays.asList(
                getClass().getResource("/tiff/quad-lzw.tif"),
                getClass().getResource("/tiff/ccitt/group3_1d.tif")
        );

        for (URL resource : testData) {
            // Read it
            BufferedImage original = ImageIO.read(resource);

            // Crop it
            BufferedImage subimage = original.getSubimage(original.getWidth() / 4, original.getHeight() / 4, original.getWidth() / 2, original.getHeight() / 2);

            // Store cropped
            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
            try (ImageOutputStream output = ImageIO.createImageOutputStream(bytes)) {
                ImageWriter imageWriter = createImageWriter();
                imageWriter.setOutput(output);
                imageWriter.write(subimage);
            }

            // Re-read cropped
            BufferedImage cropped = ImageIO.read(new ByteArrayImageInputStream(bytes.toByteArray()));

            // Compare
            assertImageEquals(String.format("Error comparing cropped to output: %s", new File(resource.toURI()).getName()), subimage, cropped);
        }
    }

    private void assertImageEquals(final String message, final BufferedImage expected, final BufferedImage actual) {
        assertNotNull(message, expected);
        assertNotNull(message, actual);
        assertEquals(message + ", widths differ", expected.getWidth(), actual.getWidth());
        assertEquals(message + ", heights differ", expected.getHeight(), actual.getHeight());

        for (int y = 0; y < expected.getHeight(); y++) {
            for (int x = 0; x < expected.getWidth(); x++) {
                assertEquals(String.format("%s, ARGB differs at (%s,%s)", message, x, y), expected.getRGB(x, y), actual.getRGB(x, y));
            }
        }
    }
````

[...]

PR #239, opened 2016-05-06
Fix for TMI #228

Schmidor:
I had to also change the dataLength calculation for uncomprompressed images as mentioned in #229, because the resized CCITT images is written as uncompressed and was written with this issue.

### DIFF
--- a/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriter.java
+++ b/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriter.java
@@ -364,8 +364,14 @@ public final class TIFFImageWriter extends ImageWriterBase {
         long stripByteCount = 0;
 
         if (compression == TIFFBaseline.COMPRESSION_NONE) {
+            int[] sampleSizes = renderedImage.getSampleModel().getSampleSize();
+            int pixelSize = 0;
+            for (int i = 0; i < sampleSizes.length; i++) {
+                pixelSize += sampleSizes[i];
+            }
+
             long ifdOffset = exifWriter.computeIFDOffsetSize(entries.values());
-            long dataLength = renderedImage.getWidth() * renderedImage.getHeight() * numComponents;
+            long dataLength = ((long) Math.ceil(renderedImage.getWidth() * pixelSize / 8.0)) * renderedImage.getHeight();
             long pointerPos = imageOutput.getStreamPosition() + dataLength + 4 + ifdOffset;
             imageOutput.writeInt((int) pointerPos);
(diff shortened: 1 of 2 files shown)


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriter.java
```diff
@@ -632,127 +637,126 @@ private void writeImageData(DataOutput stream, RenderedImage renderedImage, int
...
+        }
+
         final ByteBuffer buffer;
-        if (sampleSize == 1) {
+        if (pixelSize == 1) {
             buffer = ByteBuffer.allocate((tileWidth + 7) / 8);
         }
         else {
-            buffer = ByteBuffer.allocate(tileWidth * renderedImage.getSampleModel().getNumBands() * sampleSize / 8);
+            buffer = ByteBuffer.allocate(tileWidth * pixelSize / 8);
```
haraldk: I think this can be generalized to 

    buffer = ByteBuffer.allocate((tileWidth * pixelSize + 7) / 8);

...(ie. round up) and skip the if/else?

haraldk:
Looking mostly good!

I need to review that quite big change from moving the databuffer type switch from outside to inside the x/y loop. It could be that branch prediction saves us, but I initially wrote it like it was because I measured it to be faster. Need to do some performance testing on that. 

The upside with the code you propose is obviously that there are less duplicate blocks. :-)


Great work!

-- 
Harald K

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriter.java
Schmidor: You are right. Also this definately causes problems with pixelsizes that are not full bytes and greater that 1 bit.

Schmidor:
@haraldk There merge conflicts with the current changes because of my changes for also fixing the IFD offsets. So you either can merge this and ignore the changes from that commit or I could rebase and only add the second commit.

haraldk:
I'll manually integrate this, I need to fix the test case too, as it uses the file system and refers to `C:\temp\something` which won't work as intended on the CI server.

-- 
Harald K

Schmidor:
Oh crap, that seems again to be redundant test output which I missed to remove. The intended test is the try-block above that one.

haraldk:
@Schmidor 

While manually trying to integrate this, I found some more bugs in both the original and contributed code, and after some tinkering, I found a cleaner and easier approach (using a temporary raster for each row). 

Thanks anyway, though! :-)

-- 
Harald K

Issue #237, opened 2016-05-05
ArrayIndexOutOfBoundsException when reading G4 compressed TIFF file


THausherr:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 8
	at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decode2D(CCITTFaxDecoderStream.java:207)
	at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decodeRowType6(CCITTFaxDecoderStream.java:264)
	at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decodeRow(CCITTFaxDecoderStream.java:276)
	at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.fetch(CCITTFaxDecoderStream.java:111)
	at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.read(CCITTFaxDecoderStream.java:427)
	at java.io.DataInputStream.readFully(DataInputStream.java:195)
	at java.io.DataInputStream.readFully(DataInputStream.java:169)
	at com.twelvemonkeys.imageio.plugins.tiff.TIFFImageReader.readStripTileData(TIFFImageReader.java:1386)
	at com.twelvemonkeys.imageio.plugins.tiff.TIFFImageReader.read(TIFFImageReader.java:900)
	at javax.imageio.ImageIO.read(ImageIO.java:1448)
	at javax.imageio.ImageIO.read(ImageIO.java:1352)

```

[badccittstream.zip](https://github.com/haraldk/TwelveMonkeys/files/250130/badccittstream.zip)

(For me: the file is from PDFBOX-2188, `Root/Pages/Kids/[2]/Resources/Pattern/P2/Resources/XObject/Im1` , raw stream converted to TIF file using JavaApplication41createTIFF2 )

The TIFF file displays with IrfanView, and reads with java 7.

The problem itself appeared in our current decoder too (which is an early version of the sun decoder), as can be seen in 
https://issues.apache.org/jira/browse/PDFBOX-2187

later versions of the sun decoder increased two arrays by 1, which surpprisingly would also solve the problem for you, in CCITTFaxDecoderStream, change

```
        this.changesReferenceRow = new int[columns];
        this.changesCurrentRow = new int[columns];
```

to

```
        this.changesReferenceRow = new int[columns+1];
[...]

PR #240, opened 2016-05-06
TMI #237: Fix for CCITT images with more b/w changes than columns

Schmidor:


haraldk:
Excellent! 

Thank you!

-- 
Harald K

Issue #244, opened 2016-05-17
PSDReader incorrectly uses grayscale ICCProfile with a grayscale background.


Eurybiadan:
Hello again,

I've found a (rather specific bug) that I see in my application of the PSDReader. I'm pretty sure know exactly what the problem is, but I'm not sure how you want to approach fixing it (or I'd just submit a pull request...).

As the title suggests, the reader does not correctly use the ICC Profile of Grayscale PSD files that have a grayscale background. When a background is enabled (visible) in the attached psd (renamed as a png, of course), the intensities are washed out, consistent with a basic linear grayscale colorspace, instead of the more typical Gamma 2.2 or Dot Gain 20%.

This appears to be because the number of channels changes when there is a background. When the background is invisible (transparent), the file contains 2 channels. When the background is visible, the file contains 1 channel. This causes a variety of issues that causes the simple "ImageIO.read" to fail (i.e. load the wrong bands).

I attempted to get around this bug by creating my own ImageReadParam, but it appears that even if the data gets loaded correctly, the destination color model I choose with my ImageReadParam is not being applied (PSDImageReader:line 496) for some reason, which still causes the layers to appear washed out.

As far as fixes go, I see two options:

1) Honor every aspect of the ImageReadParam that is submitted, down to the color model, and don't convert after the fact.
2) Change lines 174 and 180 of PSDImageReader to be consistent with the other channel/bit levels, ex: `return ImageTypeSpecifiers.createBanded(cs, new int[] {0}, new int[] {0}, DataBuffer.TYPE_BYTE, false, false);`

I have workarounds like forcing the grayscale image into a more "typical" curve via LUTs (ala gamma 2.2 or Dot Gain 20%), but it would be nice if I didn't have to detect if the user has a visible background (and all the handling code that entails).

Respectfully,

RFC

![itty_bitty](https://cloud.githubusercontent.com/assets/5322065/15309015/e07a6b8a-1bb0-11e6-94c6-e079de4efe9d.png)

haraldk:
Hi Robert,

Thanks for the detailed report!

I think that ignoring the ICC profile for grayscale for single channel images is a bug/omission from my side, so we should definitively do option 2). For your test image, this seems to create colors consistent with what I see in Preview on OS X. 

[...]

PR #245, opened 2016-05-18
Fixed colorspace bug. (Issue # 244)

Eurybiadan:
As discussed in Issue #244 , I have modified the lines per option 2. It fixes the issue as expected.


haraldk:
Thanks, but there seems to be something bogus with the commit, as *all* the lines in the file are changed, according to Git... I was expecting 2 lines changed.

Not sure what is the reason for this, maybe an encoding issue?

-- 
Harald K

Eurybiadan:
Strange... I only changed those two lines. I'll see if I can fix the commit...

PR #250, opened 2016-05-20
Make the batik version configurable via a property

ankon:
This removes a bit of duplication in the POM, and also allows to set it for a build via the command line, so that testing different versions is easier.


haraldk:
Looks reasonable. Thanks.

-- 
Harald K

PR #251, opened 2016-05-20
Fallback for missing TIFFTag.PhotometricInterpretation

Schmidor:
This has the same fallback logik as JAI ImageIO.


haraldk:
Seems reasonable.

Only the samplesPerPixel == 4 to RGB mapping seems a little strange... Should probably be CMYK, unless we also have ExtraSamples. 

As always... Do you have test cases and files? :-)

I think a reasonable test would be to test (for each possible fallback) that we can load the image, and that the proper warning callback is invoked.

We really need those to avoid future regressions...

Anyway, keep up the good work!


Thanks,

-- 
Harald K

Schmidor:
Currently this is really only what JAI does.
I interpret that as "at least something is visible", independant from whether the channels are wrong or inverted. 
Do you now how an alpha channel is stored, as a normal sample or an extra sample?

Again I only have a customer file I can't share with CCITT G3 and not even that for the other cases. 
So I think I will just use some of the existing test files and strip the tag for this test case, is that ok?

haraldk:
Alpha channels should always result in an ExtraSamples tag being present, but then again... Specs and implementations... I think there's already some fallback logic in the reader. There can be more than one ExtraSamples value. But only one of them should be "associated" (premultiplied) or "non-associated" (non-premultiplied), the others should be "unspecified" (not alpha). See [ExtraSamples](http://www.awaresystems.be/imaging/tiff/tifftags/extrasamples.html) for more info. The actual samples are typically interleaved with the rest of the pixel data (or a separate plane, depending on PlanarConfiguration).

Creating test files is ok with me. I guess having one with CCITT G3 is good enough for now as well.. At least the method won't be removed without anyone noticing. Just make sure the test test for the warning.


Regards,

-- 
Harald K

Schmidor:
Ok, so including extraSamples in the check with RGB for (samplesPerPixel-extraSamples) == 3 und CMYK for 4 should be better.

Regards,
Oliver

haraldk:
Yes, I think so. :-)

-- 
Harald K

PR #255, opened 2016-05-24
Fix getNextChangingElement

Schmidor:
The Testimage is distored without this change.
lastChangchingElement must be reduced by 2 as sometimes the nextChanchingElement is before the last chanching element. Reset the last bits does not work if the number is higher.


haraldk:
I'll take your word for it. ;-)

Looks good with tests and all.

-- 
Harald K

PR #261, opened 2016-06-01
Extended AffineTransformOp for a Graphics2D fallback on filter-method

Schmidor:


haraldk:
Hmmm...

Seems the proposed changes breaks the build for some reason...? Can you look into that? :-)

Also, can you:
 *  Update the year in the BSD header to 2016 
 *  Fix the class documentation ("AffineTransformOp that uses a fallback to Graphics2D rendering if ImagingLib fails." or something)
* Use a `try/finally` around the G2D rendering, and invoke `g2d.dispose()` in the finally block
* Rename the `filter2` method, and instead `@Override` the `filter` method in AffineTransformOp
* And as always, preferably a test case too.. :-)

If you don't have time for all of the above (and I understand if you don't), I would prefer to have the change inline in TIFFUtilities (perhaps with comment that we might extract this later).


Best regards,

-- 
Harald K

Schmidor:
Unfortunately AffineTransformOp.filter is defined as final, so I can't directly overwrite it.
I found out why the tests fail: The ImagingLib checks explicitely for the calling classes and therefore rejects on derived classes.
Oh well...so the only solution seems to be to delegate to a java.awt.image.AffineTransformOp instead of extending it.

Best regards,
Oliver

haraldk:
I see. 

Well, delegation is probably cleaner than extending anyway.

-- 
Harald K

Schmidor:
With the TestCase I'm not sure what to do.
Simply running somithing that calls the new filter.method? This does not trigger the fallback. And through the ResampleOp the new AffineTransformOp is already used, which was also the cause for the failing tests in the first version.
Running through all TIFFs as I've done to demonstrate the problem would need importing the TIFF plugins and would still not guarantee to test the fallback.

Best regards,
Oliver

haraldk:
Thanks,

Looks much better now!

Re: Test case, I'll send you a starting point in a separate email. :-)

-- 
Harald K

PR #270, opened 2016-06-28
Improved area of interest

skybert:
Hi there,

We've been running with this new `com.twelvemonkeys.servlet.image.aoi.AreaOfInterest`in our own fork for years now and would love to get this into upstream so that we can ditch our own fork.

I tried to squashed the commits into one cohesive commits, but Git didn't play ball as there are literally hundreds of commits in between the commits in our fork and in upstream. The PR diff should in any case be without any fluff. 

-Torstein


skybert:
Mhmm. The diff became a lot larger than I though. Argh. I diffed against our own `master` which tracks `upstream`, but with a few exceptions. I can't find the "delete PR button" :-(

haraldk:
Hi Torstein,

Seems a lot of files were included with all lines deleted/added... Different line break or encoding setting, perhaps? I agree, we'll close this PR, and create a new one.

Let me know if I can help out in any way! :-)


Best regards,

-- 
Harald K

PR #303, opened 2016-12-15
Fix for writing TIFF pages with bogus or empty TAG_JPEG_INTERCHANGE_FORMAT_LENGTH

Schmidor:
Fixes for cases, where JpegInterChangeFormat needs to be in front of StripByteData or where JpegInterchangeFormatLength is missing.


haraldk:
Hmmmm..

I'm a little uncertain about this, as the spec dictates that tags in an IFD must be written in order (low to high). This means the StripByteCounts *MUST* always precede the JPEGInterchangeFormat tag...

Maybe there's currently a bug in the TIFFImageReader? Or am I missing something?

Also, would it be possible to change the format from old style JPEG to the new JPEG in the process? After all, the JPEG compressed data will remain unchanged...


Regards,

-- 
Harald K

Schmidor:
Hmm, the tags should still be in order, only the order of the content to which the tags point is changed.
At least if it is written as in the first file from the testcase the reader expects it that way:
https://github.com/haraldk/TwelveMonkeys/blob/master/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReader.java#L1148

I've wanted to change as less as possible in the metadata. But I assume it should be possible if InterChangeFormat and Length point at the same data as StripByteOffset/StripByteData and no JPEG Tables are contained? What about cases if there are extra JPEG Tables or differences?

Regards,
Oliver

haraldk:
Hmmm...

I see. Well... All the weird cases are pretty much all summed up here:

TIFF Technical Note 2: ftp://ftp.sgi.com/graphics/tiff/TTN2.draft.txt

TLDR; *To permit TIFF implementations to continue to read existing files, the 6.0
JPEG fields and tag values will remain reserved indefinitely.  However,
**TIFF writers are strongly discouraged from using the 6.0 JPEG design.***

Let's help make the world a better place. 😉

JPEGInterchangeFormat tags should *not* be used with JPEGxxTables tags. If they are, ignore the JPEGxxTables tags, and assume the JPEGInterchangeFormat is correct (as TIFFImageReader does). 

If the JPEGInterchangeFormatLength is bogus, it's better to just ignore it and not trying to fix it.

But I think all such files can easily be rewritten to single tile new-style JPEG without JPEGTables.

The ones with only JPEGxxTables and strips/tiles can probably just keep the strips/tiles as is, and merge the tables into a new-style JPEGTables (code for this is also in the TIFFImageReader, although it creates many full JFIF streams, instead of one table block and separate tiles).


-- 
Harald K

Schmidor:
Hmm, I really like the idea to let those who use scanners that generate JPEG-6 to suffer under their decision :P

Rewriting pages with exact one ByteStrip and InterchangeFormat offset identical to the StripByteOffset or the InterChangeFormat+Length ending directly before the StripByteOffset seems simple enough.

Am I right, that the code of "createJFIFStream" must be rewritten to work for the JPEGTables Tag?
I suppose building a complete stream with pseudo tile width/height will not work here.

Regards,
Oliver

Schmidor:
I found some information about the necessary data for JPEGTables, seems doable :)

haraldk:
Great!

Re: Suffering; Yes, I think even just stating that Old-JPEG/6 isn't supported is fine. :-) But as the reader supports these ugly beasts, it kind of makes sense. 

`JPEGTables` is used only for New-JPEG/7, but you could use most of the code from `createJFIFStream` to create the contents for it (i.e DHT/DQT). You need to add DRI and DAC handling as well according to the spec, although I've never experience these in a real-world TIFF. I'm thinking maybe we could refactor the method, to a reusable component, that creates a abbreviated JFIF stream based on a TIFF IFD, and making the contents in a format compatible for `JPEGTables`. This would also make it possible to create a unified way of reading JPEG in the reader, which would be a good thing. The current implementation (all three of them) is kind of messy...

-- 
Harald K

Schmidor:
Not supporting Old-JPEG/6 would just generate me trouble :-)

I've started generating the "abbreviated table specification" datastream, currently directely in the contrib module as I don't want to add the tiff plugin as a dependency. 

According to TIFF Technical Note 2 writers should not add DRI/DAC markers:
"Since JPEG defines the SOI marker to reset the DAC and DRI status, these
marker values cannot be carried over into any image datastream, and thus
they are effectively no-ops in the JPEGTables field.  To avoid confusion,
it is recommended that writers not place these marker types in JPEGTables.
However readers should properly skip over them if they appear."

I got a Old-JPEG/6 file that contains JPEGInterchangeFormat tags and JPEGxxxTable Tags which point into the JPEGInterchangeFormat data and has many strips with 8 rows per strip. Currently the reader cannot read it, I want to change that :-)

haraldk:
I see... 

Does that mean that they were no-ops in the Old-JPEG/6 format as well? Or how does the `JPEGRestartInterval` tag really work? Maybe we have to carry (and multiply) these tags into the "image" stream, instead of the "tables" stream.

PS: Can you send me that image file?
Also, I'm looking for a TIFF (new or old style) with JPEG lossless compression, not able to find it anywhere...

-- 
Harald K

Schmidor:
I've got no file which uses a JPEGRestartInterval > 0, so currently I would just ignore it.
For the mentioned file I assume that either ignoring the JPEGInterchangeFormat and just using JPEGTables or ignoring the tables and prepend the JPEGInterchangeFormat to each strip is the solution.

I've sent you a mail with a google drive link to the file and some lossless files with different bit depths.

haraldk:
Excellent, thank you!

.k

PR #353, opened 2017-06-07
Fix for erroneous encoded ccitt lines

Schmidor:
The ccitt tiff is missing a terminating huffman code on a line end. The following EOL is misinterpreted and the codestream read till an ArrayIndexOutOfBoundsException occurs.
Now reading the line is aborted as soon as an EOL is found.


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -356,6 +356,9 @@ private int decodeRun(final Tree tree) throws IOException {
             if (n.isLeaf) {
                 total += n.value;
                 if (n.value < 64) {
+                    if (n.value < 0) {
```
haraldk: How about testing this before the n.value < 64 branch? As 0 is < 64 anyway...

haraldk:
Excellent! 

A fix before anyone even noticed the problem was there. ;-)

I'll merge this when I find some time to work on the project again.

-- 
Harald K

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
Schmidor: I didn't want to do this check on the same level. < 64 is a normal flow for terminating codes, while < 0 is an error case. Theoretical this means the inner if is called at least a little less often for long runlengths, where non-terminating codes are used.

haraldk: Fair enough. 
I guess then we could re-arrange the branches so that the normal cases (n.value >= 64) is handled before the terminating codes, and still avoid the extra nesting?

Schmidor: I've changed the order and overwrote my commit. Is it better now? :-)

haraldk: Thumbs up! :-)

PR #368, opened 2017-07-28
An invalid IFD pointer caused creating an empty IFD Directory

Schmidor:
If the IFD pointer is pointing to an invalid location after EOF an empty directory is created by the metadata reader. This causes a wrong page count.


imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFReader.java
```diff
@@ -135,7 +135,9 @@ protected Directory readDirectory(final ImageInputStream pInput, final long pOff
                 CompoundDirectory next = (CompoundDirectory) readDirectory(pInput, nextOffset, true);
 
                 for (int i = 0; i < next.directoryCount(); i++) {
-                    ifds.add((IFD) next.getDirectory(i));
+                    if(next.getDirectory(i).size() > 0) {
```
haraldk: I'm a little reluctant to this size > 0 thing... 
Can we handle this before even trying `readDirectory`? If not, can we have `readDirectory` return an empty `CompoundDirectory` in this special case where there's really no IFD?

haraldk:
Thanks for looking into this! :-)

I made a comment on the fix, but I agree we should handle this in some way. 

-- 
Harald K

imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFReader.java
Schmidor: I did not consider there could be reasons for having empty directories.
For returning an empty `CompoundDirectory` I had to change a unit test, please look into that. 
`readSubdirectories` is creating an IFD array in this case, which caused a ClassCastException

haraldk: Didn't think of that... :-/

Maybe the best behavior would be to leave empty sub-IFDs as is, but only skip empty linked IFDs (as they make no sense) similar to what you wrote initially. Just add a comment on why this (`size == 0`) may happen, so that we can clean it up later. There's a TODO that suggests we should rewrite the method to *not* handle linked IFD parsing at all in the method, might look at both at the same time.

PR #369, opened 2017-07-28
Fix #334: Writing compressed multipage TIFF

### LINKED ISSUE #334: Writing compressed multipage TIFF
There is either a bug or need for additional clarification with the tiff-plugin when writing multiple pages is combined with compression.

https://gist.github.com/WowMuchName/23b80df8cb54917fac0b338f1e3ea245

**Expected:**

A tiff is created with two pages:

image0

image1

**Received**

A tiff with two pages:

image0

image1

**Additional information**

The pages after one seem always wrong. When using real images, they are not always completely black but always messed up with black bars all over the place. They are displayed like that in windows-image viewer and irfan view. Also reading the file-data back into java contains the same error after page one. 

When I remove the lines 34-35 (where group4 compression is set) the multitiff is okay. Using LZW instead of CCITT T.6 emitts a result which is also invalid and causes an exception when trying to read it in java.

It seems that combining multitiff + compression is either not working correctly or there is need for additional documentation. In JAI from which we are migrating I can't recall ever needing anything apart from setCompressionMode and setCompressionType to achieve the desired compression for multitiffs. If additional steps are needed in twelvemonkeys an example, testcase or additional documentation which shows how to write multitiffs with compression would be helpfull.

Schmidor:


haraldk:
Thanks! 

I already had a fix for this locally, but hadn't pushed it due to lack of tests... Do you happen to have a unit test for this? :-)

If not, don't worry, I'll make one.

-- 
Harald K

Schmidor:
No, I don't have a test. I would have added it to the commit :-)

Oliver

PR #371, opened 2017-08-03
Support PDF ByteAligned Flag on CCITT Decoder

Schmidor:


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -106,6 +111,15 @@ public CCITTFaxDecoderStream(final InputStream stream, final int columns, final
                 "CCITT GROUP 3/4 OPTION UNCOMPRESSED is not supported");
     }
 
+    /**
+     * This is used for CCITT streams from PDF files, which use EncodedByteAlign
+     *
+     * @param b EncodedByteAlign
+     */
+    public void setOptionByteAligned(boolean b) {
```
haraldk: Can this value change while decoding? Otherwise, I'd prefer the field to be final, and set in the constructor only.  And a proper parameter name. :-)

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -241,11 +255,12 @@ private int getNextChangingElement(final int a0, final boolean white) {
     }
 
     private void decodeRowType2() throws IOException {
-        resetBuffer();
+        if(optionByteAligned) resetBuffer();
```
haraldk: Code style: Always use curly brackets, and space after `if`.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -241,11 +255,12 @@ private int getNextChangingElement(final int a0, final boolean white) {
...
         decode1D();
     }
 
     private void decodeRowType4() throws IOException {
+        if(optionByteAligned) resetBuffer();
```
haraldk: Code style: Always use curly brackets, and space after `if`.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -272,6 +287,7 @@ private void decodeRowType4() throws IOException {
     }
 
     private void decodeRowType6() throws IOException {
+        if(optionByteAligned) resetBuffer();
```
haraldk: Code style: Always use curly brackets, and space after `if`.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -366,17 +382,7 @@ private int decodeRun(final Tree tree) throws IOException {
...
-        }
-
-        while (true) {
-            if (bufferPos == -1) {
-                return;
-            }
-
-            readBit();
-        }
+        bufferPos = -1;
```
haraldk: This change looks a bit suspicious... 
Though it could just as well be that the previous code wasn't needed. Which leads me to... Do you have a unit test? :-)

Schmidor: This code effectively never needed:
readBit() also just counts up bufferPos und returned the current bool.
Reseting decodedRow also is done in the filling loop starting on line 311.
I'm sure the present unit tests would have flagged it if it changed behaviour.
I'm just now realized this code is redundant.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
Schmidor: It cannot change while decoding. For normal CCITT decoding as in TIFF compressions it should never be changed from the default: false for G3/G4, true for MODIFIED HUFFMAN RLE. Changing this flag from default is only a special case for PDF CCITT, which is why I would not like it as an additional constructor parameter.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
Schmidor: Unfortunately, I have exactly one PDF which sets EncodedByteAlign and I can't share it.

haraldk: Would it be possible to create an "artificial" sample stream for a test case? As usual, I'm not worried that the current implementation is correct. It's just that I can change it without breaking any tests. This is bound to happen sooner or later as part of some refactoring...

haraldk:
Excellent!

PR #373, opened 2017-08-22
Support Modified and custom tags

diorcety:
Allows to customize tags in the generated tiff file. It allows, for example, to modify only the image of an existing tiff image with custom fields.

### DIFF
--- a/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageMetadata.java
+++ b/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageMetadata.java
@@ -1338,4 +1338,31 @@ public final class TIFFImageMetadata extends AbstractMetadata {
     public Entry getTIFFField(final int tagNumber) {
         return ifd.getEntryById(tagNumber);
     }
+
+    public Iterable<Entry> getTIFFFields() {
+        return new Iterable<Entry>() {
+            @Override
+            public Iterator<Entry> iterator() {
+                return new Iterator<Entry>() {
+
+                    private final Iterator<Entry> iterator = ifd.iterator();
+
+                    @Override
+                    public boolean hasNext() {
+                        return iterator.hasNext();
+                    }
+
+                    @Override
+                    public Entry next() {
+                        return iterator.next();
+                    }
+
+                    @Override
+                    public void remove() {
+
+                    }
+                };
+            }
+        };
+    }
 }
--- a/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriteParam.java
+++ b/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriteParam.java
@@ -28,8 +28,12 @@
 
 package com.twelvemonkeys.imageio.plugins.tiff;
 
+import com.twelvemonkeys.imageio.metadata.Entry;
+
 import javax.imageio.ImageWriteParam;
+import java.util.HashMap;
 import java.util.Locale;
+import java.util.Map;
 
 /**
  * TIFFImageWriteParam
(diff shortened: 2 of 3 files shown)


haraldk:
Hi @diorcety !

Thank you for your contribution! Much appreciated!

I do understand your use case I think. However the standard way of passing existing or custom TIFF fields in the ImageIO API, is through the metadata object. This is currently only partially supported, but I think we should stick to the standard, rather than inventing a new, non-standard way. Sorry about that. You should probably file a bug for missing support for writing custom fields. :-)

If you want to have a go at this, please have a look at the `com.twelvemonkeys.imageio.plugins.tiff.TIFFImageWriter#writePage` method, there's a TODO there:

        // TODO: There shouldn't be necessary to create a separate map here, this should be handled in the
        // convertImageMetadata/getDefaultImageMetadata methods....

The idea is that the above mentioned methods should create a *full* set of metadata, based on the input metadata or created from scratch, that will be written to the IFD for the current image. 

The `convertImageMetadata` should just pass on any custom tags it does not understand or care about (like comments, copyright, etc). Some baseline tags needs to be stripped/changed depending on the new image data to be written (like compression, photometric, width, height, tiles/strips etc), this is currently partially handled. 

Finally, to make sure the implementation is correct, and more importantly, stays correct, I will need one or more test cases (as new `@Test` annotated methods in the `TIFFImageWriterTest` class) with the PR.


Best regards,

-- 
Harald K

diorcety:
I wanted to use that system, the issue is that the library inject custom stuff. My goal is to load tiff, edit one pixel save the tiff and the only difference in the saved tiff compared to the loaded one if this pixel.

haraldk:
As I said, we can support your *use case*, but I don't want to add a custom API for this. Instead, we fix the implementation so that the standard API works as it should, and preserves tags in the metadata. 

No "custom stuff" should be injected. But some of the tags must be re-created from the image data, so that the tags match the image data that is written, otherwise the TIFF will not be readable (ie. the image may be have been cropped/scaled, and different encoders may use different compression/tile size/strip size etc.). One can also argue that updating the Software tag is appropriate, but I think it will currently only be updated if there's no value present in the metadata.

If we want to retain strip/tile size, extra work is needed in the writer, but that is possible. The compression may not write exactly the same data though, so that offsets and lengths will still change unless the data is uncompressed. 

-- 
Harald K

diorcety:
I am agree on that. When I talked about only one byte modification on the saved tiff compare to the input, of course I talked about uncompressed data. For example with that PR (except this part which seems good: https://github.com/haraldk/TwelveMonkeys/pull/373/commits/71fce60646e9f4bf2eca4eb985b82edef15dae1a#diff-2c4522629e22d7a79f77618db41a9675R473)
I need to remove these attributes which are not in my input file, and are not mandatory (or the default value is the good value)
            TIFF.TAG_ICC_PROFILE,
            TIFF.TAG_RESOLUTION_UNIT,
            TIFF.TAG_SAMPLES_PER_PIXEL,
            TIFF.TAG_SOFTWARE,
            TIFF.TAG_ORIENTATION,
            TIFF.TAG_X_RESOLUTION,
            TIFF.TAG_Y_RESOLUTION

A good test should be to do a test with a binary diff between an input file and the saved file without any modification.

haraldk:
Sorry if I sound mean, but no, the idea of binary diff equality between two TIFF files isn't going to fly... 

Even if we write the exact same tags, with the exact same values and the same pixel data, two files written by different encoders ([or even different versions of the same encoder](https://stackoverflow.com/q/46790556/1428606)) (link about PNG, but same issues apply) will most likely be binary different. The thing is that the TIFF specification allows many ways to mix tags, "tag values" for tags > 4 bytes and the image data (and even other, non-TIFF data if you like). Trying to emulate every other library out there will be practically impossible. 

So the only thing that makes sense to me, is to aim for:

1: Image data equality (the same dimensions, same pixel values)
2: Meta data equality (the same tags with the same values, as much as possible)

Even the second point will be quite hard, because it's not only "meta" data, it's really image layout parameters. So you would need to add support for writing images with arbitrary strip/tile sizes from the original image. And again, any offset values will probably not be the same, because the writer may decide to write tags/tag data before or after (or in-between!) image tiles/strips.

So I think your best bet would be to test for image data equality, and filter out a subset of the tags you need to retain, and test for equality on those.


Best regards,


-- 
Harald K

PR #380, opened 2017-09-13
IOCA plugin

mattcg:
IOCA is a format which is used to contain image data, typically within AFP files, although it may occur outside of them.

This plugin allows MODCA (AFP) parsers to delegate the parsing of an IOCA stream to it.

In turn, the plugin delegates parsing of JPEG-compressed or TIFF compressed data to the an ImageReader loaded at runtime.

For G4 and G3-encoded fax streams, it uses the CCITT decoder provided by the TIFF plugin.

Based on the [IOCA 6.1 reference](http://afpcinc.org/wp-content/uploads/2016/08/IOCA-Reference-Image-Object-Content-Architecture-Reference.pdf).


haraldk:
Hi Matthew,

Thank you for this great contribution! I just had a quick look at the code so far, and it looks quite impressive!

I will try to merge this into the project soon, but I've got my hands full now with cleaning things up for the 3.4 release. Mostly bug fixes will go into that release. IOCA support will go into the next major release, along with other new features.  

I see the build failed, but only Java 7, so it may be a Travis thing, but great if you have time to look inot it! 


Again, thanks! 

-- 
Harald K

mattcg:
Thank you for the positive initial review :) It seems the build failed due to a problem with the environment.

While we're on the subject, d'you have plans to deprecate support for Java 7? It would make the code much simpler! Lambdas would allow to refactor to very concise code that uses the strategy pattern.

haraldk:
From a developer standpoint I'd love to switch to Java 9 right away, as all the Java 8 features would make the code better and less verbose, and Java 9 would make quite a few things simpler. 

But I still get issues asking for support for Java 6 (or even 1.5!). So it's not that easy to just make the switch. We'll have to stay with Java 7 for the time being. Unfortunately. :-/

It's not really a problem implementing a strategy pattern in Java 7 though, you just can't use the new APIs or the nice lambda syntax.

-- 
Harald K

mattcg:
Understood, and in the latest pushes I implemented in a Java 7 way in fact.

haraldk:
Hi @mattcg! 

Sorry about my complete lack of attention to this old PR, that I even said I would merge. 😞 
My apologies!

Over the years, though, there hasn't been any requests for adding this, so I believe the demand just isn't there. Still, it might be a useful addition to some, so you could consider releasing it as a separate package. 

I'm closing this PR for now.

PR #417, opened 2018-04-04
Support for 16bit floating point data type in TIFF

ikhaldeev:
Allow reading 16-bit floating point data type.

Hacky implementation based on SO answer to conversion half-precision float.


haraldk:
Thank you, Ivan!

This looks good! The build errored for some reason, but I think it's a temporary Travis issue.

I'll look into it in more detail soon, and hopefully your patch will make it into the 3.5 release (3.4 is very near completion, and long overdue, so I prefer not to destabilize it with more features right now). 

Again, thank you for the contribution!


Best regards,

-- 
Harald K

haraldk:
Sorry for waiting so long... Due to the failing check, I had to apply this manually, but now the fix is in! Thank you! 

-- 
Harald K

PR #420, opened 2018-04-06
Failing test for profile header issue

ikhaldeev:
Here is failing test for the created [issue](https://github.com/haraldk/TwelveMonkeys/issues/421). Unfortunately due to implementation details of Profiles caching it is hard to create a test that would fail during running full test suite, so in order to reproduce a bug you could run only this new test.

`mvn -Dtest=ColorSpacesTest#testCMYKFromProfileAlwaysSame -DfailIfNoTests=false test`


haraldk:
Thank you for pointing out the bug and providing a way to reproduce!

I will close this PR without merging, as I prefer tests that don't rely on modifying the file system. Running the the current test will still find the issue on systems with an installed CMYK profile. Not perfect, but I think good enough for now. 


Best regards,

-- 
Harald K

PR #429, opened 2018-06-29
Fix getAlphaRaster for ExtraSamplesColorModel (TIFFImageReader)

ikhaldeev:
Default `getAlphaRaster` method from `ComponentColorModel` uses an incorrect band for alpha channel in case there are some unspecified extra components. 

Here are fix and test for it.


haraldk:
Looking good! Thanks!

-- 
Harald K

PR #431, opened 2018-07-16
Added an extra mark to match the reset call in the finally block

seafraf:
The finally block is always called, so previously this was calling reset twice and marking only once.

This caused an issue with some of my dependencies that had set marks on the stream before `canDecodeInput` was called, this would eat the marks and when their code called reset it would not do was it was expected.


haraldk:
Thanks @seafraf !

I'll look into this as soon as possible (just got home from vacation). Do you also have a test case that would fail with the old code, and pass with your fix, to verify and prevent regressions?

-- 
Harald K

PR #432, opened 2018-07-27
Fix ResampleOp for ExtraSamplesColorModel

ikhaldeev:
The default way of computing significant bits in ComponentColorModel does not take into account extra components. This leads to ArrayIndexOutOfBoundsException in `getComponentSize` method when trying to resample image with ResampleOp#filter.

Don't know whether this fix is conceptually correct. Does it make sense to resample extra components?
At least provided test illustrates the issue.


haraldk:
Thanks Ivan! Looks good. 

I just came home from 3 weeks vacation, but I'll review and merge as soon as I find the time!

-- 
Harald K

haraldk:
Thanks! I used your test case, expanded on it, and implemented a similar fix in the color model.

PR #436, opened 2018-08-21
TIFFUtilities: Enhanced support for old-style to new-style JPEG transformation

Schmidor:
I got some more old-stlye-JPEG TIFFs that were corrupted on merging or splitting pages, which are handled now,

As there are many duplications from the TIFFReader for type 6 there is still a bit of refactoring to be done, such that the contrib module can directly use the JPEGTables/stream generating methods from the reader.


haraldk:
Looking good. :-)

PR #446, opened 2018-09-25
Fix for corrupted output from non-RGB TIFF images with extra samples

bernhardf-ro:
Non-PHOTOMETRIC_RGB TIFF images with non-alpha extra samples result in corrupted output.

Adding a check to create an ImageTypeSpecifier with an ExtraSamplesColorModel solved the issue.
For a more universal and cleaner solution I added a method that centralizes the use of "createBanded" and "createInterleaved" as well as the "extra samples" special case for all applicable interpretations.


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReader.java
```diff
@@ -701,6 +661,18 @@ else if (bitsPerSample <= 0 || bitsPerSample > 16) {
         }
     }
 
+    private ImageTypeSpecifier create(int planarConfiguration, ColorSpace cs, int dataType, int significantSamples, int samplesPerPixel, boolean alpha, boolean alphaPremultiplied) {
```
haraldk: I kind of like this generalization. But the method should probably be named `createImageTypeSpecifier`.
Also, your change breaks the existing logic in several places where the original code would fall through and throw exception (ie. if someone introduced a new value for the `PlanarConfiguration` field), your new method will create a banded spec, which is incorrect, and will probably fail at some random place later.
I would prefer if you kept the original swich logic in this method, and added a `default` that simply throws an exception about the unsupported planar config.

bernhardf-ro:
Changed the method according to your comment.

Sorry, but the image I spotted the issue with is from an issue report by a customer. So I cannot publish it.
If anyone wants to create a test image: It is a CMYK TIFF with 1 extra channel that is not tagged as alpha (neither associated nor unassociated)

haraldk:
Thanks! This looks good.

I understand, it's a common situation. Can you ask the customer to release just this one (or a similar) image for use in testing? If not, I'll try to craft one. 


In any case, thanks a lot,

-- 
Harald K

bernhardf-ro:
I added an artificially created test file. (Converted an existing cmyk sample to TIFF with alpha and hex-edited the extra-channel tag.)

PR #465, opened 2019-01-01
Fix test that doesn't compile when the JDK is Java 11 and remove a few tests that are only testing java.util classes.

jimfcarroll:
Fix test that doesn't compile when the JDK is Java 11 and above and remove a few tests that are only testing java.util classes.

The few tests removed seem that they are no longer testing custom collection classes and are now simply testing java.util collections.

### DIFF
--- a/common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionAbstractTest.java
+++ b/common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionAbstractTest.java
@@ -1136,7 +1136,7 @@ public abstract class CollectionAbstractTest extends ObjectAbstractTest {
         verifyAll();
 
         try {
-            array = collection.toArray(null);
+            array = collection.toArray((Object[])null);
             fail("toArray(null) should raise NPE");
         } catch (NullPointerException e) {
             // expected
--- a/common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionAbstractTest.java
+++ b/common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionAbstractTest.java
@@ -1136,7 +1136,7 @@ public abstract class CollectionAbstractTest extends ObjectAbstractTest {
         verifyAll();
 
         try {
-            array = collection.toArray((Object[])null);
+            array = collection.toArray((Object[]) null);
             fail("toArray(null) should raise NPE");
         } catch (NullPointerException e) {
             // expected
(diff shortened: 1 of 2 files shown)


common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionAbstractTest.java
```diff
@@ -1136,7 +1136,7 @@ public void testCollectionToArray2() {
         verifyAll();
 
         try {
-            array = collection.toArray(null);
+            array = collection.toArray((Object[])null);
```
haraldk: This change looks ok. Please insert a space between the cast and the `null` to follow code style.

common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionUtilTest.java
```diff
@@ -228,34 +228,12 @@ public void testArrayListIteratorRange() {
         assertCorrectListIterator(CollectionUtil.iterator(new String[] {"foo", "bar", "baz", "boo"}, 1, 2), new String[] {"bar", "baz"});
     }
 
-    @Test
```
haraldk: Why did you remove these tests? I don't see anything wrong with these?

haraldk:
I believe the problem you face is due to a new method in JDK 11, not JDK 8. But I guess it won't hurt making things forward compatible. :-)

common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionUtilTest.java
jimfcarroll: I could be wrong but it appears these tests don't test any of your code. The methods `assertCorrectListIterator` (which I made static just to be sure they didn't depend on the state of something set on the test class) have no dependencies on your custom collection classes and nothing passed into them in these test do either. It appears your testing `java.util.Arrays.asList`. 

Originally I thought all of your tests in this section did this but I see now you've created a test bed for you collection extentions. Did I miss something in this case?

common/common-lang/src/test/java/com/twelvemonkeys/util/CollectionAbstractTest.java
jimfcarroll: Okay. I added the space. I can squash the commits eventually if you'd prefer. Just let me know once I get it right. Thanks.

haraldk:
Manually submitted the important parts of this PR, leaving the rest as-is.

PR #485, opened 2019-05-13
Fix JPEG component ID array size reference

eeflores:
Changed the size value to be in line with the defined array size in `ComponentIdSet`.

Unfortunately I can't provide a sample that was causing the issue, but the JPEG image in question was malformed in some way. The software using this library was caught waiting for the iteration through the `int` range - appeared to be frozen.

Changing the expected size here "fixed" the problem for me, however rewriting the (int) id value back to the byte buffer may be a potential problem?


haraldk:
Hi @eeflores ,

Thanks for the PR! 

I'll look into it!

I agree with you that the current value of `<= 16` makes absolutely no sense, as the size can never grow larger than 4... But I'm not sure (can't remember) what the initial idea was. Perhaps it should instead be `id < 255`? Most likely it won't matter much, as the read will probably fail anyway in this case... :-)


-- 
Harald K

eeflores:
Hi Harald,

Thanks for the feedback! If I get some time maybe I can learn more about what's happening under the hood.

Regards,

Erwin Flores.

haraldk:
I believe this was fixed via #510 and will close now.

PR #495, opened 2019-08-22
Svg scale by ViewBox

Schmidor:
When a SVG only contains a ViewBox like "0 0 100 100" instead of width/height definitions, a default size of 400 / 400 is used.
This results in wrong image scaling if  SVGReadParam.setSourceRenderSize is used.


haraldk:
Looks good,

PR #505, opened 2019-10-22
Changes suggested by Intellij in the category "Performance".

KoenDG:
The main change here is in TIFFImageReader.java

Once it has found that the condition has been met once, there is no point in continuing the loop, as the entire purpose was to determine this value for the very next step after the loop and no external variables have been changed.

### DIFF
--- a/common/common-lang/src/main/java/com/twelvemonkeys/lang/BeanUtil.java
+++ b/common/common-lang/src/main/java/com/twelvemonkeys/lang/BeanUtil.java
@@ -88,7 +88,7 @@ public final class BeanUtil {
 
         while (begIdx < pProperty.length() && begIdx >= 0) {
 
-            endIdx = pProperty.indexOf(".", endIdx + 1);
+            endIdx = pProperty.indexOf('.', endIdx + 1);
             if (endIdx > 0) {
                 subProp = pProperty.substring(begIdx, endIdx);
                 begIdx = endIdx + 1;
@@ -106,7 +106,7 @@ public final class BeanUtil {
             Class[] paramClass = new Class[0];
 
             int begBracket;
-            if ((begBracket = subProp.indexOf("[")) > 0) {
+            if ((begBracket = subProp.indexOf('[')) > 0) {
                 // An error if there is no matching bracket
                 if (!subProp.endsWith("]")) {
                     return null;
--- a/common/common-lang/src/main/java/com/twelvemonkeys/util/regex/WildcardStringParser.java
+++ b/common/common-lang/src/main/java/com/twelvemonkeys/util/regex/WildcardStringParser.java
@@ -622,7 +622,7 @@ public class WildcardStringParser {
             buffer.append("\n");
             buffer.append("      Format: <state index>: <character> (<last free state>)");
             buffer.append("\n");
-            buffer.append("      Number of strings parsed: " + totalNumberOfStringsParsed);
+            buffer.append("      Number of strings parsed: ").append(totalNumberOfStringsParsed);
             buffer.append("\n");
         }
         return buffer.toString();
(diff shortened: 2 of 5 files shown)


common/common-lang/src/main/java/com/twelvemonkeys/lang/BeanUtil.java
```diff
@@ -106,7 +106,7 @@ public static Object getPropertyValue(Object pObject, String pProperty) {
             Class[] paramClass = new Class[0];
 
             int begBracket;
-            if ((begBracket = subProp.indexOf("[")) > 0) {
+            if ((begBracket = subProp.indexOf('[')) > 0) {
```
KoenDG: indexOf on char should be ever so slightly faster than on String. Less instructions, smaller object. Really small stuff though. Still, why not.

common/common-lang/src/main/java/com/twelvemonkeys/util/regex/WildcardStringParser.java
```diff
@@ -622,7 +622,7 @@ public String toString() {
             buffer.append("\n");
             buffer.append("      Format: <state index>: <character> (<last free state>)");
             buffer.append("\n");
-            buffer.append("      Number of strings parsed: " + totalNumberOfStringsParsed);
+            buffer.append("      Number of strings parsed: ").append(totalNumberOfStringsParsed);
```
KoenDG: Doing a regular string `+` inside a stringbuilder append defeats the purpose of having the stringbuilder to begin with. This is the only example found of this.

haraldk:
Hi Koen,

Thanks for your pull request! All performance improvements, even minor ones, are appreciated. 

Unfortunately, IntelliJ seems to be in error about the change in `TIFFImageReader`. If you read the code, you'll see that the loop actually builds the color map. I also do a quick check in the same loop, to detect an 8 bit color map to avoid looping twice. But that's not the main purpose of the loop, so adding a break there, will... break the code (I see I'm missing a test case for this, so I should probably add one).


Best regards,

-- 
Harald K

KoenDG:
Oh damn, looks like I had the situation reversed. I was under the impression that the entire array got re-done in the second loop if the condition in the first was met, but it's exactly the opposite.

imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReader.java
```diff
@@ -196,7 +197,7 @@ public int getHeight(int pImageIndex) throws IOException {
         checkBounds(pImageIndex);
 
         // TODO: Better implementation, include INT_RGB types for 3BYTE_BGR and 4BYTE_ABGR for INT_ARGB
-        return Arrays.asList(getRawImageType(pImageIndex)).iterator();
+        return Collections.singletonList(getRawImageType(pImageIndex)).iterator();
```
KoenDG: Since the return of `getRawImageType` is but a single item, a singletonlist can be used.

As the documentation states: `Returns an immutable list containing only the specified object.`

Should be smaller in memory than the default arraylist. It's apparently used a few more times along the codebase.

KoenDG:
Code updated

haraldk:
@KoenDG 

That's probably just a local problem on your side. The project currently must be built with Oracle Java 7 or 8, for the tests to pass (due to hard coded color values in some tests not matching the output of the new color management system, usually). As the code compiles and test fine on Travis/CI, it's good as far as I'm concerned.

-- 
Harald K

imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReader.java
haraldk: The TODO above the changed line indicates the reason for using a mutable list here... But of course, we could change this until we actually return multiple values.

KoenDG:
Indeed, I'm running Amazon's Corretto 8 on the machine I'm testing from.

imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReader.java
KoenDG: Right. Change removed.

PR #509, opened 2019-11-24
Fix for PICT files on Mac

thilo-bauer:
I created a fix by adjusting the buffer sizes in the PICT image reader. Otherwise files copies via macOS clipboard often cannot be rendered. In pre-Catalina versions this could also affect files copied from macOS Preview app. This is not an issue with Catalina, because pictures transferred now are PNG format instead. However, certain CAD apps on macOS may still rely on exchange of old PICT format via clipboard transfer. I hope, this fix is welcome.

### DIFF
--- a/imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
+++ b/imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
@@ -78,7 +78,6 @@ import java.awt.image.*;
 import java.io.*;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 
@@ -182,7 +181,7 @@ public final class PICTImageReader extends ImageReaderBase {
             readPICTHeader0(pStream);
         }
     }
-
+    
     private void readPICTHeader0(final ImageInputStream pStream) throws IOException {
         // Get size
         picSize = pStream.readUnsignedShort();
@@ -326,7 +325,7 @@ public final class PICTImageReader extends ImageReaderBase {
 
     /**
      * Reads the PICT stream.
-     * The contents of the stream will be drawn onto the supplied graphics
+     * The contents of the stream will be drawn onto the supplied graphics 
      * object.
      * <p>
      * If "DEBUG" is true, the elements read are listed on stdout.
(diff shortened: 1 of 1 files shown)


haraldk:
Hi Thilo,

Thanks for your suggestions! I will look through them in more detail when I find the time.

In the mean time, it would be nice if you could add some test cases containing clipboard data, that would fail using the original code and pass with your fix. Otherwise, there's no way to know if this fixes anything. And a refactoring may break it again, without anyone noticing.

Best regards,

-- 
Harald K

imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
```diff
@@ -2058,10 +2057,10 @@ else if (packType == 4) {
         int[] pixArray = null;
         short[] shortArray = null;
         if (packType == 3) {
-            shortArray = new short[srcRect.height * (rowBytes + 1) / 2];
+            shortArray = new short[srcRect.height * (srcRect.width + 1)];
```
haraldk: I don't think the +1 here is correct, unless you're also missing a / 2...? It was used in the original calculation to get the integer division truncation to "round" up.

imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
```diff
@@ -2058,10 +2057,10 @@ else if (packType == 4) {
...
         }
         else {
-            pixArray = new int[srcRect.height * (rowBytes + 3) / 4];
+            pixArray = new int[srcRect.height * (srcRect.width + 3)];
```
haraldk: I don't think the +3 here is correct, unless you're also missing a / 4...? It was used in the original calculation to get the integer division truncation to "round" up.

imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
```diff
@@ -2628,7 +2627,7 @@ public int getHeight(int pIndex) throws IOException {
 
     public Iterator<ImageTypeSpecifier> getImageTypes(int pIndex) throws IOException {
         // TODO: The images look slightly different in Preview.. Could indicate the color space is wrong...
-        return Collections.singletonList(
+        return Arrays.asList(
```
haraldk: You should probably keep the `Collections.singletonList` here, unless you have a  specific reason to change it?

haraldk:
Closing this for now, as I haven't been able to verify the changes. If you can provide test files (clipboard dumps) that will not be read without these changes, but will work with them, feel free to reopen!

-- 
Harald K

PR #517, opened 2020-01-28
Minor code cleanup: Intellij suggested changes from static code analysis.

KoenDG:
I had Intellij's static code analysis have a run over the code. Not much to find that could be improved without touching the logic of the code heavily, this is minor cleanup.

### DIFF
--- a/common/common-lang/src/main/java/com/twelvemonkeys/util/Time.java
+++ b/common/common-lang/src/main/java/com/twelvemonkeys/util/Time.java
@@ -129,8 +129,7 @@ public class Time {
      * @see #toString(String)
      */
     public String toString() {
-        return "" + getMinutes() + ":"
-                + (getSeconds() < 10 ? "0" : "") + getSeconds();
+        return getMinutes() + ":" + (getSeconds() < 10 ? "0" : "") + getSeconds();
     }
 
     /**
--- a/common/common-lang/src/main/java/com/twelvemonkeys/util/TimeFormat.java
+++ b/common/common-lang/src/main/java/com/twelvemonkeys/util/TimeFormat.java
@@ -44,12 +44,12 @@ import java.util.Vector;
  * The format is expressed in a string as follows:
  * <DL>
  * <DD>m (or any multiple of m's)
- * <DT>the minutes part (padded with 0's, if number has less digits than 
+ * <DT>the minutes part (padded with 0's, if number has less digits than
  *     the number of m's)
  *     m -&gt; 0,1,...,59,60,61,...
  *     mm -&gt; 00,01,...,59,60,61,...
  * <DD>s or ss
- * <DT>the seconds part (padded with 0's, if number has less digits than 
+ * <DT>the seconds part (padded with 0's, if number has less digits than
  *     the number of s's)
  *     s -&gt; 0,1,...,59
  *     ss -&gt; 00,01,...,59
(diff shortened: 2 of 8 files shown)


common/common-lang/src/main/java/com/twelvemonkeys/util/Time.java
```diff
@@ -129,8 +129,7 @@ public int getMinutes() {
      * @see #toString(String)
      */
     public String toString() {
-        return "" + getMinutes() + ":"
-                + (getSeconds() < 10 ? "0" : "") + getSeconds();
+        return getMinutes() + ":" + (getSeconds() < 10 ? "0" : "") + getSeconds();
```
KoenDG: The starting `"" + ` is not at all necessary.

common/common-lang/src/main/java/com/twelvemonkeys/util/TimeFormat.java
```diff
@@ -44,12 +44,12 @@
  * The format is expressed in a string as follows:
  * <DL>
  * <DD>m (or any multiple of m's)
- * <DT>the minutes part (padded with 0's, if number has less digits than 
+ * <DT>the minutes part (padded with 0's, if number has less digits than
```
KoenDG: It looks like stray spaces were picked up in the comments.

common/common-lang/src/main/java/com/twelvemonkeys/util/TimeFormat.java
```diff
@@ -404,7 +404,7 @@ String format(Time t) {
 
         // Else return it with leading 0's
 	//return StringUtil.formatNumber(t.getSeconds(), digits);
-        return StringUtil.pad("" + t.getSeconds(), digits, "0", true);
+        return StringUtil.pad(String.valueOf(t.getSeconds()), digits, "0", true);
```
KoenDG: Actual change. Using `String.valueOf()` instead of concatenating an empty string.

The code analysis states that For the purpose of turning something into a String, concatenating an empty string is never going to be as good performing as `String.valueOf()`

contrib/src/main/java/com/twelvemonkeys/contrib/tiff/TIFFUtilities.java
```diff
@@ -200,11 +200,10 @@ public static void rotatePage(ImageInputStream imageInput, ImageOutputStream ima
     }
 
     public static List<TIFFPage> getPages(ImageInputStream imageInput) throws IOException {
-        ArrayList<TIFFPage> pages = new ArrayList<TIFFPage>();
-
         CompoundDirectory IFDs = (CompoundDirectory) new TIFFReader().read(imageInput);
 
-        int pageCount = IFDs.directoryCount();
+        final int pageCount = IFDs.directoryCount();
+        List<TIFFPage> pages = new ArrayList<>(pageCount);
```
KoenDG: In this particular case, it possible to reliably determine the exact size that will be used.

common/common-lang/src/main/java/com/twelvemonkeys/util/TimeFormat.java
haraldk: `Integer.toString` might be even faster in this case, skipping one indirection... But I guess this would still be an improvement. :-)

Issue #526, opened 2020-02-18
Preventing SSRF due to external resource refs in SVGs


actinium15:
Until batik 1.13.0-SNAPSHOT it is not possible to prevent SSRF vulnerability exploits due to SVG external resources. TwelveMonkey's `imageio-batik`, which relies on batik 1.9 at the moment, suffers from this. See the test file as linked at [0].

Apache Batik fixed it via [BATIK-1276](https://issues.apache.org/jira/browse/BATIK-1276), and this issue is to track rolling-up this change in TwelveMonkeys as soon as new batik version is released.

Very trivially, if external-resource processing is to be blocked altogether in SVG processing for TwelveMonkeys' `imageio-batik`, a patch like at [1] should be sufficient.
While we wait for [Batik's update](https://issues.apache.org/jira/browse/BATIK-1276?focusedCommentId=17038799&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17038799), I'd like to invite the opinion of project maintainers on sufficiency of changeset at [1] from `imageio-batik`'s functionality perspective. If I receive guidance on this I'd be happy to work on a patch as soon as a batik release with [BATIK-1276](https://issues.apache.org/jira/browse/BATIK-1276) is publically available.

[0] https://issues.apache.org/jira/secure/attachment/12988316/test.svg
[1]
```
diff --git a/imageio/imageio-batik/pom.xml b/imageio/imageio-batik/pom.xml
index 385f666..3dacf15 100644
--- a/imageio/imageio-batik/pom.xml
+++ b/imageio/imageio-batik/pom.xml
@@ -88,6 +88,6 @@
     </dependencies>

     <properties>
-        <batik.version>1.9</batik.version>
+        <batik.version>1.13.0-SNAPSHOT</batik.version>
     </properties>
 </project>
diff --git a/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java b/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
index c55012d..96f2060 100755
--- a/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
+++ b/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
[...]

PR #529, opened 2020-02-25
#526 Preventing SSRF due to external resource refs in SVGs

actinium15:


actinium15:
hi @haraldk, may I request you to provide guidance on the changeset? I'd be happy to work on improving it should you have any suggestions... Thanks!

haraldk:
Sorry, haven't had the time to look into it yet. I will! 😀

-- 
Harald K

imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
```diff
@@ -80,6 +80,7 @@
  */
 public class SVGImageReader extends ImageReaderBase {
     private Rasterizer rasterizer;
+    private boolean allowExternalResources;
```
haraldk: Just thought about it, and we could do something like:

```suggestion
    private boolean allowExternalResources = "true".equalsIgnoreCase(System.getProperty("com.twelvemonkeys.imageio.plugins.svg.allowexternalresources"));
```

haraldk:
Thanks, I was about to write that we could do this much simpler, but you beat me to it. 😀

The new version is nearly there. However I have some changes I would like to do:

1. Instead of the constant for the system property name, I would like to have a *single* constant for the *global setting* (ie: `final static boolean DEFAULT_ALLOW_EXTERNAL_RESOURCES = "true".equalsIgnoreCase(System.getProperty(...));` (as we now read the system property only once, we no longer need a constant for its name, just keep the string inline). This will fix the setting to the *initial* value of the system property, to avoid confusing behaviour if someone tries to change it at runtime. This is also more secure.
2. Unfortunately, this also affects our tests. But in reality, this is also a good thing, because test may run in parallell, and relying on setting/clearing system properties could potentially result in flaky tests. The tests that relies on external resources will have to be read using param.
3. Both the local settings (reader and param) will be initialized to the value of the global setting (as now, just the constant instead of the system property). 

With these changes, I hope to be able to merge your PR. 😀

Great work so far!

-- 
Harald K

actinium15:
Thanks for your guidance @haraldk!

I've pushed another commit based on your latest review (I hope I got all the things you wanted right). Please do let me know in case I've missed something. Thanks for your understanding and apologies in case I've changed something the way you didn't expect (esp the part where I enabled the system-property for whole of the test-JVM in the pom - per my understanding that's the only way to be certain for reliable parallel execution of tests, though that did mean I could no longer assert the default (no system-property, no API invocation) behaviour in the test)

Issue #416, opened 2018-04-03
TIFFImageWriter ignores custom TIFF tags


matthieu-labas:
Custom TIFF tags (i.e. not in "baseline" or "extended") are ignored in TIFFWriter.

	BufferedImage bi = new BufferedImage(64, 64, BufferedImage.TYPE_BYTE_GRAY);
	byte[] raster = ((DataBufferByte)image.getRaster().getDataBuffer()).getData();
	Arrays.fill(raster, (byte)0xaa);
	TIFFImageWriter writer = (TIFFImageWriter)ImageIO.getImageWritersByFormatName("TIFF").next();
	try (ImageOutputStream os = ImageIO.createImageOutputStream("/tmp/test.tiff")) {
		writer.setOutput(os);
		List<Entry> head = new ArrayList<>();
		head.add(new TIFFEntry(TIFF.TAG_PHOTOMETRIC_INTERPRETATION, TIFF.TYPE_SHORT, 0));
		head.add(new TIFFEntry(1000, TIFF.TYPE_ASCII, "Toto")); // <<-- Custom tag
		TIFFImageMetadata header = new TIFFImageMetadata(head);
		TIFFImageWriteParam param = (TIFFImageWriteParam)writer.getDefaultWriteParam();
		IIOImage image = new IIOImage(bi, null, header);
		writer.write(null, image, param);
	} finally {
		writer.dispose();
	}

A TIFF dump (or hex file dump) shows no `"Toto"` in the resulting file.

matthieu-labas:
In `TIFFImageWriter`, the `entries` map is populated with TIFF fields *queried* from the image metadata (correcting them if needed).

It works if we initialize that map with the initial IFD after creating the map (`TIFFImageWriter.java:180`):

    Map<Integer, Entry> entries = new LinkedHashMap<>();
    for (Iterator<Entry> iter = metadata.getIFD().iterator(); iter.hasNext(); ) {
    	Entry ent = iter.next();
    	entries.put((Integer)ent.getIdentifier(), ent);
    }

haraldk:
Hi Matthieu,

Thanks for reporting! 

I'm aware of the issue (this is basically the same issue as #412), only a specific subset of TIFF tags are carried over into the resulting TIFF file.

[...]

PR #554, opened 2020-08-03
Removed custom metadata entry (tag) restrictions

githabibi:
Custom tags are still being removed as part of the code for re-creating the metadata in the convertImageMetadata method. By removing the checks that "vet" the tags, custom tags maybe entered without issues.

### DIFF
--- a/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriter.java
+++ b/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageWriter.java
@@ -771,40 +771,7 @@ public final class TIFFImageWriter extends ImageWriterBase {
         if (ifd != null) {
             for (Entry entry : ifd) {
                 int tagId = (Integer) entry.getIdentifier();
-
-                switch (tagId) {
-                    // Baseline
-                    case TIFF.TAG_SUBFILE_TYPE:
-                    case TIFF.TAG_OLD_SUBFILE_TYPE:
-                    case TIFF.TAG_IMAGE_DESCRIPTION:
-                    case TIFF.TAG_MAKE:
-                    case TIFF.TAG_MODEL:
-                    case TIFF.TAG_ORIENTATION:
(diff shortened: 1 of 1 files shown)


githabibi:
This is in reference to #416 which a fix was claimed, despite custom tags still not being written.

haraldk:
Hi githabibi,

Thank you for the PR! 

I am absolutely interested in allowing more custom tags. However, I am reluctant to implement the changes you suggested, as it makes it possible to create corrupted TIFFs/illegal combinations of tags. This is the reason for the current, somewhat restrictive approach. When I implemented it, I had the choice between allowing all tags I know was okay, or remove all tags I knew was not okay. And chose the former to be safe. 

Perhaps if you described your actual use case, we could work out something that would allow you to write the tags you want, while still making sure we create "good" TIFFs. 😀


Best regards,

-- 
Harald K

githabibi:
Hey Harald,

     The use case I'm interested in is one where I'm writing a completely custom tag. Similar to #416 except the range that I'm writing is 10,000 - 15,000.

     Nothing interesting there, just working on legacy software that was previously using JAI and ImageIO, which the latter I know was successful in writing those custom tags; but I've found your library has been absolutely amazing, well-written and error-averse, so I definitely prefer to use it. I'd really like to avoid some crazy solution such as registering the classes on the fly in order to be able to write the tags. 

Thank you!

haraldk:
Ok,

According to [Wikipedia](https://en.wikipedia.org/wiki/TIFF#Private_tags) TIFF has a "private" tag id range (32,768 and above), and a "reusable" tag id range (65,000–65,535) that developers can as they see fit. It does seem that your range is also "okay" (as in "no other known TIFF tags in this range"), although using the 65000+ range is probably safer.

I think we ca find a solution, but we need to make sure we don't allow writing corrupted/illegal TIFFs. 


-- 
Harald K

githabibi:
Alright, well, the issue is due to the nature of the code I'm working on, the tag key must be in that range. Perhaps I edit this PR such that it allows the writing of tags within that range?

haraldk:
Fixed via 9c89770. 
I relaxed the constraints quite a bit, to allow most tags above 1000.

PR #580, opened 2021-01-04
Use GitHub CI to build project on major versions of Java and Linux, MacOS & Windows

carldea:
I hope this project will like this PR. This is a GitHub CI workflow that has four runners to build & test concurrently on LTS versions of Java and on the major OSes. Any pushes to master will trigger the job to occur.

### DIFF
--- a/.github/workflows/ci-build-test.yml
+++ b/.github/workflows/ci-build-test.yml
@@ -1,0 +1,30 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+---
+name: Java CI
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, macOS-latest, windows-2016]
+        java: [1.8, 11, 15, 16-ea]
+      fail-fast: false
+      max-parallel: 4
+    name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+
+      - name: Verify with Maven
+        run: mvn verify -B --file pom.xml
+        env:
+          MAVEN_OPTS: -Dsun.java2d.cmm=sun.java2d.cmm.kcms.KcmsServiceProvider 
+...
--- a/.github/workflows/ci-build-test.yml
+++ b/.github/workflows/ci-build-test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-18.04, macOS-latest, windows-2016]
-        java: [1.8, 11, 15, 16-ea]
+        java: [1.8, 11, 15]
       fail-fast: false
       max-parallel: 4
     name: Test JDK ${{ matrix.java }}, ${{ matrix.os }}


haraldk:
Thanks!

I've been thinking about changing to GitHub CI (from the current Travis CI integration) for a while, so this is certainly welcome!

Some questions: 

1. On Travis we currently run a 4 build matrix, on Oracle JDK 1.8 with KCMS (the `-Dsun.java2d.cmm=sun.java2d.cmm.kcms.KcmsServiceProvider` option), on Oracle JDK 1.8 with no settings, and with no settings on Oracle JDK 11 & 15. From what I understand, the GitHub workflow you propose will create 9 builds (3 OS x 3 JDK versions, all with the KCMS option), am I right? I need to test Java 1.8 both with and without this option (for later versions it no longer applies). 
2. The current CI runs for PRs as well as pushes to any branch. Can this be configured the same way? (I probably don't need the full matrix for PR or non-master branches). 
3. Do you know what JDK builds that run on the various OSs? Is it all Oracle JDK?
4. Finally, it would be cool to have a "latest" JDK build that only triggered a warning if it failed, without breaking the build. Is this possible? (this can always be added later, It's just something I've been wanting to add)

I will look more into this later! Again, thanks for doing this! 👍 


-- 
Harald K

carldea:
Harald,
Glad to help! Sorry for the late reply.

> 1.  On Travis we currently run a 4 build matrix, on Oracle JDK 1.8 with KCMS (the -Dsun.java2d.cmm=sun.java2d.cmm.kcms.KcmsServiceProvider option), on Oracle JDK 1.8 with no settings, and with no settings on Oracle JDK 11 & 15.

I will have to do some more research and get back to you in regards to the property (sun.java2d.cmm).
> 1. ... From what I understand, the GitHub workflow you propose will create 9 builds (3 OS x 3 JDK versions, all with the KCMS option), am I right?

Yes.

> 1 ... I need to test Java 1.8 both with and without this option (for later versions it no longer applies).

I will have to do some more research and get back to you in regards to the property (sun.java2d.cmm).

Either way, within GitHub Ci it's easy to break out jobs (separate), with or without the property. I'm not sure what versions when/if they no longer apply.
 
> 2. The current CI runs for PRs as well as pushes to any branch. Can this be configured the same way? (I probably don't need the full matrix for PR or non-master branches).

Yes. I can update the workflow to do that. 

You can do the following:
```yaml
on: [push]
```
or 

```yaml
on:
  push:
    branches:    
      - master
      - 'beta/myspecial_feature'
```

> 3. Do you know what JDK builds that run on the various OSs? Is it all Oracle JDK?

Under the hood, GitHub Actions actions/setup-java uses the Zulu builds of OpenJDK. Zulu has a consistent track record of carrying the widest range of OpenJDK versions (6,7,8,9,10,11,12,13,14,15, as well as e.g. 16-ea and 17-ea right now) in a consistently accesible CDN that has not flaked, moved, or broken in ways that break CI/CD tests over time. ;-) 

> 4. Finally, it would be cool to have a "latest" JDK build that only triggered a warning if it failed, without breaking the build. Is this possible? (this can always be added later, It's just something I've been wanting to add)

As a proposed best-practice in general, I would recommend having matrixes that contain both a fixed version and the latest update in each Java version.
For example:
```yaml
java: [7, 7.0.181, 8, 8.0.192, 9, 10, 11, 11.0.3, 12, 13, 13.0.4, 14, 15, 16-ea, 17-ea]
```
The logic is that you want to be using fixed update levels for your CI/CD tests, so that when things break it will be clear that it's your code and not the jdk update that broke them. You also want to use the latest update levels so that when new JDK updates come out that break things, you'll know about them. Having both makes sense to do, and lets you quickly triage problems when they are detected.

I only pick fixed update versiosn for "stable" JDK versions. E.g. I consider  7, 8, 11, and 13 to be stable, in the sense that they were not "just recently released" and are still being updated. For EOL'ed versions (9, 10, 12, 14), using just the latest is fine, as no new updates will be coming, so it's a fixed version. And for "very young" versions (15 right now) I also skip using a fixed version. Tho that last one may arguably have 15.0.1 added as a fixed version.

The list can obviously be pruned over time. Libraries probably benefit from testing on a wide raneg of versions and keeping even the "orphaned" ones around, but other projects can probably skip e.g. 9,10,12,14, and use the latest flavor of the 6 month feature releases (e.g. 16 when it comes out) but drop them a year or two later.

Please let me know if you want me to change the code for this PR or a separate PR.

Thank you,
-Carl

haraldk:
Hi again Carl,

Thanks! And sorry for my late reply, I just don't have much time to follow up at the moment. 

I will need to test with the Kcms system property, but as this applies only to Java 1.6/1.7/1.8 it sounds like the best option perhaps is to just run one extra job with 1.8 and the property set.

I think PR builds are special, in the they actually need to do a check out of the target branch, then merge the source branch and finally build the result of that. It's not just a matter of building a single branch. I would assume GitHub CI has functionality for that? 

Your proposed list of JDKs is quite long, but it does sound nice to have a full check on all known versions of Java... As long as GitHub has no limit on this I'm okay with that. It would also be nice as I said, to have latest versions of Java only give warnings, not fail the build (again, to keep the build stable).

Anyway, really appreciate this PR! 😀 

-- 
Harald K

carldea:
Harald,
Sure no problem, I got busy too. 

Yes, let me know how you want to proceed. If you want you can implement how you want things done (gleaning from my PR) and just close the PR. 

GitHub CI allows you to separate jobs, via the yml file or separating it entirely in another yml file sitting in the workflows directory.

Carl

gotson:
Let me see how we can test Oracle JDK 8 with KCMS on Github Workflow.

For 4.x, you should move to the new LTS, which is 17. Just my opinion, but that is for instance what Spring is going to do for version 6.x.

haraldk:
Fixed via #633

PR #599, opened 2021-03-26
Allow embedded resource URLs if external resources are disabled

Schmidor:
Currently if _allowExternalResources_ is _false_ resources embedded with "data:" URLs are disabled also.
The EmbededExternalResourceSecurity accepts those and no other.


haraldk:
Sounds like a good change! Thanks! 😀 

-- 
Harald K

Issue #579, opened 2020-12-23
Broken Display of TIFF with CCITT Group 3 Fax Encoding


GuidoRahn:
We recently updated our application to Java 11. It is a JavaFx application in which Images in various image formats / compressions can be displayed.

There were some problems with the display of images and images in PDF documents (displayed via PDFBox). By updating to the latest available version of the twelvemonkey library 3.6.1, one of our customers encountered a new problem with certain Tiff image files with 'CCITT Group 3 Fax Encoding'.
These Tiff image files are displayed distorted and cut off.

If we go back to our previous used version 3.4.1, the Tiff image files are displayed correctly.

Attached is a sample-Application, which shows how we display images in our application and which also includes an imagefile in the resources-directory. The Image is loaded on Click on the 'Load Image'-Button.
[FaxViewer.zip](https://github.com/haraldk/TwelveMonkeys/files/5733780/FaxViewer.zip)

haraldk:
Hi Guido,

Thanks for reporting! 

I'm unfortunately not able to look at the source code for your JavaFX application. But `resources/sample/Fax.tiff` seems to be show the problem, and I can see the problem right away.

We introduced a way to discover a different set TIFF files that had the encoding set incorrectly, that looks at the CCITT stream to discover the presence of leading EOLs, and if not found, reverts to RLE encoding. Unfortunately, your file is in Group 3/T.4 format, but is still missing the EOLs... So it's decoded "incorrectly". 

These kinds of problems typically occurs with CCITT streams extracted from PDF files and put into TIFF containers without properly re-encoding. So if possible, have a look at the software that created the file, and see if something can be done. But we'll look into it!


Best regards,

-- 
Harald K

haraldk:
@Schmidor I'm kind of hoping you have clever fix for this! 😉 Any input is appreciated though. Is leading EOLs required by the spec for G3/T.4?

haraldk:
Okay,

I think actually the check is still good, we just need to test a few more bytes for the EOL code. Changing from 20 to 32 seems to be enough. @Schmidor Can this be done smarter? 

I have a fix soon. 

-- 
Harald K

Schmidor:
@haraldk 
I believe the case for the detection here was, that the leading EOL was missing, but there were EOLs between the lines. Searching a few more bytes shouldn't be a problem. If simply increasing the data helps, it just seems the first line is just longer.

[...]

Issue #535, opened 2020-03-23
Problem with T.4 group 3 fax TIFFs


JaxCoder:
I have a bunch of mulitpage TIFF files taken from a healthcare system and when I try to display them, they are pure white. I've attempted to research TIFF compression and debug and tried making some changes, but alas, I have been unsuccessful thus far. 

the tiffinfo -d command shows a premature EOL,  and I believe this might be the issue in reading it, however - I know these TIFF files are not corrupt (at least not in the sense they are unreadable) because Snowbound java library reads them and .net framework's tiff decoder reads them as well.

Unfortunately I can't upload a TIFF here because it contains personal health information. I'll include as much info as I can, and i'll continue looking for a TIFF that is safe to upload, but im hoping someone will know why these TIFFs aren't working and maybe have an idea how to fix.

@Schmidor  it looks like you created the original G3/4 decoder, not sure if you have time to look into this.

I've included just the IFD0 info for brevity.

**TIFFINFO -D**
TIFFReadDirectory: Warning, Unknown field with tag 293 (0x125) encountered.
TIFF Directory at offset 0x8 (8)
  Subfile Type: multi-page document (2 = 0x2)
  Image Width: 1700 Image Length: 2200
  Resolution: 200, 200 pixels/inch
  Bits/Sample: 1
  Compression Scheme: CCITT Group 3
  Photometric Interpretation: min-is-white
  FillOrder: msb-to-lsb
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 1
  Rows/Strip: 32
  Planar Configuration: single image plane
  Page Number: 0-0
  Group 3 Options: EOL padding (4 = 0x4)
Fax3Decode1D: Warning, Premature EOL at line 0 of strip 0 (got 0, expected 1700).

**TIFFDUMP**
Magic: 0x4949 <little-endian> Version: 0x2a <ClassicTIFF>
Directory 0: offset 8 (0x8) next 67710 (0x1087e)
SubFileType (254) LONG (4) 1<2>
ImageWidth (256) LONG (4) 1<1700>
ImageLength (257) LONG (4) 1<2200>
BitsPerSample (258) SHORT (3) 1<1>
Compression (259) SHORT (3) 1<3>
Photometric (262) SHORT (3) 1<0>
FillOrder (266) SHORT (3) 1<1>
[...]

PR #608, opened 2021-05-27
#579 Deeper EOL search in the CCITT stream

Schmidor:
I've got other TIFFs which have some junk at the beginning of the CCITT stream, which is even longer than before. In an extrem case almost 300 bytes.

So I propose to increase the read limit for the CCITT type detection even further, but only read on demand till a EOL is found or the limit is reached.

Also relates to #535

### DIFF
--- a/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
+++ b/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
@@ -174,13 +174,11 @@ final class CCITTFaxDecoderStream extends FilterInputStream {
                 for (int i = 12; i < limitBits; i++) {
                     if (i % 8 == 0) {
                         read = in.read();
-                        if (read > -1) {
-                            streamByte = (byte) read;
-                        }
-                        else {
+                        if (read == -1) {
                             // no EOL before stream end
                             return TIFFBaseline.COMPRESSION_CCITT_MODIFIED_HUFFMAN_RLE;
                         }
+                        streamByte = (byte) read;
                     }
 
                     b = (short) ((b << 1) + ((streamByte >> (7 - (i % 8))) & 0x01));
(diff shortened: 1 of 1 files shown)


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -152,39 +152,50 @@ public CCITTFaxDecoderStream(final InputStream stream, final int columns, final
...
+                    return type;
+                }
+                short b = (short) (((((byte) first) << 8) + ((byte) second)) >> 4);
+                int limitBits = limit * 8;
+                int read = second;
+                byte streamByte = (byte) read;
+                for (int i = 12; i < limitBits; i++) {
+                    if (i % 8 == 0) {
+                        read = in.read();
+                        if (read > -1) {
```
haraldk, line 177: Invert the `if`, then you won't need the `else`, for better readability.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -152,39 +152,50 @@ public CCITTFaxDecoderStream(final InputStream stream, final int columns, final
...
+                    // correct, starts with EOL or byte aligned EOL
+                    return type;
+                }
+                short b = (short) (((((byte) first) << 8) + ((byte) second)) >> 4);
+                int limitBits = limit * 8;
+                int read = second;
+                byte streamByte = (byte) read;
+                for (int i = 12; i < limitBits; i++) {
+                    if (i % 8 == 0) {
+                        read = in.read();
```
haraldk, line 176: Using as many as 500 single-byte `read`s over one array `read` might impact performance negatively, but I guess it's not a deal breaker.

But would it be hard to read blocks of bytes instead?

haraldk (changes requested):
Looks mostly good, thanks! 👍🏻 

I'd just like one change to aid readability. 

Also, I would be very happy if you also added a sample file that would fail with the current logic and test, to avoid regressions. Perhaps you can just update the current test/replace the current sample file. 

Great work! Keep it up! 😀

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
Schmidor: Done :)

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
haraldk: Fair enough! 😀

haraldk (approved):

PR #612, opened 2021-06-27
Documentation cleanup

KoenDG:
Greetings.

I was setting up a new IDE and was testing the static code analysis.

This popped up and I felt like making a PR for it:

Added the `@Deprecated` tag to instances where is was mentioned in documentation, but not for the actual code itself.

Changed one documentation link pointing at a non-existing item.

### DIFF
--- a/servlet/src/main/java/com/twelvemonkeys/servlet/image/ImageFilter.java
+++ b/servlet/src/main/java/com/twelvemonkeys/servlet/image/ImageFilter.java
@@ -146,7 +146,7 @@ public abstract class ImageFilter extends GenericFilter {
      * @param pRequest the original request
      * @return the new response, or {@code pResponse} if the response is already wrapped
      *
-     * @see com.twelvemonkeys.servlet.image.ImageServletResponse
+     * @see com.twelvemonkeys.servlet.image.ImageServletResponseImpl
      */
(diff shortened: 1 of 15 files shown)


servlet/src/main/java/com/twelvemonkeys/servlet/image/ImageFilter.java
```diff
@@ -146,7 +146,7 @@ protected void doFilterImpl(final ServletRequest pRequest, final ServletResponse
      * @param pRequest the original request
      * @return the new response, or {@code pResponse} if the response is already wrapped
      *
-     * @see com.twelvemonkeys.servlet.image.ImageServletResponseWrapper
```
KoenDG, line 149: Seems like this doesn't actually exist.

haraldk:
Thanks Koen,

Most of this code has been around since before Java had annotations... But I guess it won't hurt to add these, even if the code is otherwise unmaintained.

I believe the missing class should simply be `ImageServletResponseImpl` (probably renamed at some point in time, pre GitHub migration).

KoenDG:
"Most of this code has been around since before Java had annotations..."


I read that and my mouth went agape :P

Missing class renamed, as per your remark.

Shall I squash or does github support merge with squash?

haraldk:
Thanks!

Issue #626, opened 2021-09-16
Found a failing TIFF file


KoenDG:
**Describe the bug**
A TIFF file is failing to be read.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.
3.7.0

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 
```
openjdk version "1.8.0_222"
OpenJDK Runtime Environment Corretto-8.222.10.1 (build 1.8.0_222-b10)
OpenJDK 64-Bit Server VM Corretto-8.222.10.1 (build 25.222-b10, mixed mode)
```

I added the file on my local branch, it can be found there, I accidentally made a PR first: https://github.com/haraldk/TwelveMonkeys/pull/625

The file: https://github.com/KoenDG/TwelveMonkeys/blob/1031364e6ca4ef4126837d99e154e3cd22b8075c/imageio/imageio-tiff/src/test/resources/bigtiff/G32DS.tif

**Expected behavior**
The file is processed normally.

**Stak trace**
```
java.io.IOException: Unknown code in Huffman RLE stream
        at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decodeRun(CCITTFaxDecoderStream.java:445)
        at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decode1D(CCITTFaxDecoderStream.java:227)
        at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decodeRowType2(CCITTFaxDecoderStream.java:331)
        at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.decodeRow(CCITTFaxDecoderStream.java:373)
        at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.fetch(CCITTFaxDecoderStream.java:198)
        at com.twelvemonkeys.imageio.plugins.tiff.CCITTFaxDecoderStream.read(CCITTFaxDecoderStream.java:524)
        at java.io.DataInputStream.readFully(DataInputStream.java:195)
        at java.io.DataInputStream.readFully(DataInputStream.java:169)
        at com.twelvemonkeys.imageio.plugins.tiff.TIFFImageReader.readStripTileData(TIFFImageReader.java:1878)
        at com.twelvemonkeys.imageio.plugins.tiff.TIFFImageReader.read(TIFFImageReader.java:1102)
```

[...]

PR #627, opened 2021-09-16
#626: Respect fillOrder in CCITT findCompressionType

Schmidor:
#626


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -150,15 +150,21 @@ public CCITTFaxDecoderStream(final InputStream stream, final int columns, final
     }
 
     static int findCompressionType(final int encodedType, final InputStream stream) throws IOException {
+        return findCompressionType(encodedType, stream, TIFFBaseline.FILL_LEFT_TO_RIGHT);
+    }
```
haraldk, line 154: I don't see the need for keeping the old method. It's just used in tests. You can inline it in the tests, for easy migration.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -150,15 +150,21 @@ public CCITTFaxDecoderStream(final InputStream stream, final int columns, final
...
         if (encodedType == TIFFExtension.COMPRESSION_CCITT_T4 && stream.markSupported()) {
             int limit = 512;
-
             try {
                 stream.mark(limit);
-
-                int first = stream.read();
-                int second = stream.read();
+                byte first = (byte) stream.read();
+                byte second = (byte) stream.read();
```
haraldk, line 163: I don't think it's safe to cast these values to `byte` right away, as you'll mask the `-1`/EOF value. The test for `-1` later on will also potentially match for `0xFF`...

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -199,6 +206,16 @@ else if (first == 0 && (((byte) second) >> 4 == 1 || ((byte) second) == 1)) {
...
+    private static byte reverse(byte x) {
+        byte b = 0;
+        for (int i = 0; i < 8; ++i) {
+            b <<= 1;
+            b |= (x & 1);
+            x >>= 1;
+        }
+        return b;
+    }
+
```
haraldk, line 218: Looks correct to me, but we already have a reverse byte lookup table in `ReverseInputStream`. It's probably faster. 😀 

Come to think of it, why don't we just wrap the stream when creating CCITT decoder streams like we do for other decoders? See `TIFFImageReader#createDecompressorStream`.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
```diff
@@ -167,18 +173,19 @@ else if (first == 0 && (((byte) second) >> 4 == 1 || ((byte) second) == 1)) {
...
                 int limitBits = limit * 8;
-                int read = second;
-                byte streamByte = (byte) read;
+                byte streamByte = second;
                 for (int i = 12; i < limitBits; i++) {
                     if (i % 8 == 0) {
-                        read = stream.read();
-                        if (read == -1) {
+                        streamByte = (byte)stream.read();
+                        if (streamByte == -1) {
```
haraldk, line 182: Same as above, this will interpret a `0xFF` "real" byte as EOF.

imageio/imageio-tiff/src/test/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReaderTest.java
```diff
@@ -939,6 +939,16 @@ public void testStreamMetadataMM() throws IOException {
...
+    @Test
+    public void testCCITTTypeDetectionFillOrderRightToLeft() throws IOException {
+        ImageReader reader = createReader();
+
+        try (ImageInputStream stream = ImageIO.createImageInputStream(getClassLoaderResource("/tiff/ccitt/G32DS.tif"))) {
+            reader.setInput(stream);
+            reader.read(0);
+        }
+    }
+
```
haraldk, line 951: 👍🏻 
Could probably also just add it to the test file set at the top, but this is fine for now. 😀

haraldk (changes requested):
THANKS! 👍🏻 

Looks mostly good, but needs a few changes:

* The casting of `stream.read()` to `byte` may result in incorrect behavior. See inline comments.
* I think we could potentially simplify things by using the `ReverseInputStream` , instead of handling fill order in the CCITT decoder.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/CCITTFaxDecoderStream.java
Schmidor: Good idea, I moved the fillOrder handling outside to the reader.
I think the ReverseInputStream didn't exist back when I implemented the CCITT decoder 😃

haraldk (approved):
👍🏻  This seems to have addressed all my concerns.

haraldk:
@ianbondoc The answer is, as always: As soon as I find the time. If you think I should spend more time making releases, you can help out by becoming a [sponsor](https://github.com/sponsors/haraldk).

PR #648, opened 2021-12-19
Support Group Layer in psd

younseunghyun:
Hi 

Thank you for creating TwelveMonkeys library .

I use this library with appreciation.

Recently, I have to support a group attribute at PSD.

PSD has a group, and the group visible attribute decide child layers have to drawn or not when convert to other format Image.

So I write some code for that.

Thanks


imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -33,7 +33,7 @@
 import javax.imageio.IIOException;
 import javax.imageio.stream.ImageInputStream;
 import java.io.IOException;
-import java.util.Arrays;
+import java.util.*;
```
haraldk, line 36: This file does not follow the code style. Please reformat the code according the project's code style.

* Prefer single class imports
* Use space after keywords (`if`, `for`, `while` etc)
* Use space before curly brace
* Use `else` on separate line

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -57,14 +57,22 @@
     private String unicodeLayerName;
     private int layerId;
 
+    static private Stack<List<PSDLayerInfo>> groupStack = new Stack<>();
+    static private List<PSDLayerInfo> group = null;
```
haraldk, line 61: I believe using `static` here is an error.

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDMetadata.java
```diff
@@ -387,6 +387,10 @@ private Node createLayerInfoNode() {
             node.setAttribute("left", String.valueOf(psdLayerInfo.left));
             node.setAttribute("bottom", String.valueOf(psdLayerInfo.bottom));
             node.setAttribute("right", String.valueOf(psdLayerInfo.right));
+            node.setAttribute("layerId", String.valueOf(psdLayerInfo.getLayerId()));
+            node.setAttribute("groupLayerId", String.valueOf(psdLayerInfo.groupLayerId));
+            node.setAttribute("isGroup", String.valueOf(psdLayerInfo.isGroup));
+            node.setAttribute("isSectionDivider", String.valueOf(psdLayerInfo.isSectionDivider));
```
haraldk, line 393: Perhaps these attributes could default to the most common value, and be set only if different from the default?

haraldk, line 393: These attributes also needs to be added to the `PSDMetadataFormat`.

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -144,7 +153,15 @@
...
                     break;
-
+                case PSD.lsct:
+                    sectionDividerSettingType = SectionDividerSetting.fromInt(pInput.readInt());
+                    if (resourceLength >= 16) {
+                        pInput.skipBytes(12);
+                    }
+                    else if (resourceLength >= 12) {
+                        pInput.skipBytes(8);
+                    }
```
haraldk, line 163: I believe the following would be both simpler to understand and more correct (as we read 4 bytes in the `readInt()` method):

```suggestion
                    pInput.skipBytes(resourceLength - 4);
```

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -161,6 +178,30 @@
...
+                info.groupLayerId = this.layerId;
+            }
+            group = groupStack.pop();
+            group.add(this);
+            this.isGroup = true;
+        } else {
+            if(group != null){
+                group.add(this);
+            }
+        }
```
haraldk, line 203: Please reformat this code block using the correct code style.

Also, the usage of `static` variables will not work if reading multiple PSD images at the same time.

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -191,4 +232,22 @@ public String toString() {
...
+        public static SectionDividerSetting fromInt(int value){
+            for(SectionDividerSetting rt : SectionDividerSetting.values()){
+                if(rt.value == value){
+                    return rt;
+                }
+            }
+
+            return null;
+        }
+    }
```
haraldk, line 252: This block has code style issues.

In addition to the above, use one enum constant per line.

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -191,4 +232,22 @@ public String toString() {
         builder.append("]");
         return builder.toString();
     }
+
+    enum SectionDividerSetting {
+        LAYER(0), OPEN_FOLDER(1), CLOSED_FOLDER(2), BOUNDING_SECTION_DIVIDER(3);
+        SectionDividerSetting(int value) { this.value = value;}
+
+        private final int value;
```
haraldk, line 240: Why do we need the `value`? It seems to always equal the enum `ordinal`.

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -191,4 +232,22 @@ public String toString() {
...
+        public int value() { return value; }
+
+        public static SectionDividerSetting fromInt(int value){
```
haraldk, line 243: I prefer: 

```suggestion
        public static SectionDividerSetting valueOf(int value){
```

imageio/imageio-psd/src/test/java/com/twelvemonkeys/imageio/plugins/psd/PSDImageReaderTest.java
```diff
@@ -104,7 +104,9 @@ protected ImageReaderSpi createProvider() {
                 // CMYK, uncompressed + contains some uncommon MeSa (instead of 8BIM) resource blocks
                 new TestData(getClassLoaderResource("/psd/fruit-cmyk-MeSa-resource.psd"), new Dimension(400, 191)),
                 // 3 channel, RGB, 32 bit samples
-                new TestData(getClassLoaderResource("/psd/32bit5x5.psd"), new Dimension(5, 5))
+                new TestData(getClassLoaderResource("/psd/32bit5x5.psd"), new Dimension(5, 5)),
+                // group layer psd file
+                new TestData(getClassLoaderResource("/psd/group_layer.psd"), new Dimension(4000, 1600))
```
haraldk, line 109: Unless this PSD file contains special features that needs general tests, you don't need to add it here. The specific test below will do! 😀

imageio/imageio-psd/src/test/java/com/twelvemonkeys/imageio/plugins/psd/PSDImageReaderTest.java
```diff
@@ -532,4 +534,41 @@ public void testReadUnicodeLayerName() throws IOException {
             assertEquals("If_The_Layer_Name_Is_Really_Long_Oh_No_What_Do_I_Do", ((IIOMetadataNode) layerInfo.item(0)).getAttribute("name"));
         }
     }
+
+    @Test
+    public void testGroupLayerRead() throws IOException {
+        PSDImageReader imageReader = createReader();
+
+        try (ImageInputStream stream = ImageIO.createImageInputStream(getClassLoaderResource("/psd/group_layer.psd"))) {
```
haraldk, line 542: Can you please provide a smaller test file? Both in image dimensions and file size. For the test, a 16x16 image would suffice and save about 7 MB of data.

imageio/imageio-psd/src/test/java/com/twelvemonkeys/imageio/plugins/psd/PSDImageReaderTest.java
```diff
@@ -532,4 +534,41 @@ public void testReadUnicodeLayerName() throws IOException {
...
+                    assertEquals(layerInfo.isSectionDivider, false);
+                }
+
+                // green_bag layer divider
+                if(layerInfo.getLayerId() == 43){
+                    assertEquals(layerInfo.groupLayerId == null, true);
+                    assertEquals(layerInfo.isGroup, false);
+                    assertEquals(layerInfo.isSectionDivider, true);
+                }
+            }
```
haraldk, line 571: This block has code style issues.

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
```diff
@@ -57,14 +57,22 @@
     private String unicodeLayerName;
     private int layerId;
 
+    static private Stack<List<PSDLayerInfo>> groupStack = new Stack<>();
+    static private List<PSDLayerInfo> group = null;
+    Integer groupLayerId;
+    Boolean isGroup = false;
+    Boolean isSectionDivider = false;
```
haraldk, line 64: Is there a reason these are boxed types? I usually prefer using primitives when possible.

haraldk (changes requested):
Hi! Thank you for your contribution! 

Having layer groups will be a nice addition! 👍🏻 

I have looked through your changes, and have some comments that needs to be resolved before merging with the library. 

* The major issue here is the use of `static` in the `PSDLayerInfo` class. I believe this will cause threading issues if multiple instances are used concurrently (which can happen).
* Another important thing is code style (I described the most important in the comments inline). 

Hoping you will find the time to resolve these!


Thanks, 

Harald K

imageio/imageio-psd/src/test/java/com/twelvemonkeys/imageio/plugins/psd/PSDImageReaderTest.java
```diff
@@ -532,4 +534,41 @@ public void testReadUnicodeLayerName() throws IOException {
...
+    public void testGroupLayerRead() throws IOException {
+        PSDImageReader imageReader = createReader();
+
+        try (ImageInputStream stream = ImageIO.createImageInputStream(getClassLoaderResource("/psd/group_layer.psd"))) {
+            imageReader.setInput(stream);
+
+            IIOMetadata metadata = imageReader.getImageMetadata(0);
+
+            List<PSDLayerInfo> layerInfos = ((PSDMetadata) metadata).layerInfo;
+
```
haraldk, line 548: Another thing: You need to assert that this list is indeed 3 items and contains the ids 40, 42 and 43, otherwise the loop below will simply pass with nothing being tested. Which is probably not the intention. 😀

younseunghyun:
Follow comments
Add a comment for some points(enum ordinal)

imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSDLayerInfo.java
younseunghyun: ordinal can used as value of SectionDividerSetting.

But I prefer use value when enum derived from the value.

Because, a position of enum can be changed then it can't stand the value.

So I prefer using value.

If the code style is use ordinal then value, i will follow the rule  :)

haraldk:
Thanks a lot! This looks better! I will need to perform a proper review another day, but I think this looks fine. 

I generally agree with you on the enum values. If the constants are specified in the PSD format, it may make sense. And if possible, keep the enum private or package protected (not public).

younseunghyun:
> Thanks a lot! This looks better! I will need to perform a proper review another day, but I think this looks fine.
> 
> I generally agree with you on the enum values. If the constants are specified in the PSD format, it may make sense. And if possible, keep the enum private or package protected (not public).

Okay I will make the enum as private.

I will apply the change with a your next review :)

haraldk (approved):
👍🏻

haraldk:
@younseunghyun Do you know if Photoshop really supports nested layer groups? If so, we need a sample that contains nested layer groups + tests. 

But for now, I have simplified the logic a lot, and it works fine for all the test data in the repository, so I assume that's fine.

younseunghyun:
@haraldk yes Photoshop support nested layer group.

I will test the case and add some example.

PR #650, opened 2021-12-23
Added changed from ab13084f448e01ea939847aade21c3c3e060ef98 as fallba…

koenr-bc:
Commit ab13084f448e01ea939847aade21c3c3e060ef98 breaks OSGi support due to classloader issues (what happens is Class.forName in com.twelvemonkeys.imageio.util.IIOUtil#lookupProviderByName throws a ClassNotFoundException, using the lookup as fallback works)

### DIFF
--- a/imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReaderSpi.java
+++ b/imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReaderSpi.java
@@ -39,6 +39,7 @@ import javax.imageio.stream.ImageInputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.util.Iterator;
 import java.util.Locale;
 
 import static com.twelvemonkeys.imageio.util.IIOUtil.lookupProviderByName;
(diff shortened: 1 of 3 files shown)


haraldk:
Hi @koenr-bc,

Thanks for reporting!

However, I don't like to have all that duplicated code reverted. If the code doesn't work, it's better to fix it at the source, I think.

How about instead changing the method `IIOUtil.lookuProviderByName` to:

```java
    public static <T> T lookupProviderByName(final ServiceRegistry registry, final String providerClassName, Class<T> category) {
        Iterator<T> providers = registry.getServiceProviders(category, true);

        while (providers.hasNext()) {
            T provider = providers.next();

            if (provider.getClass().getName().equals(providerClassName)) {
                return provider;
            }
        }

        return null;
    }
```

This way we avoid the `Class.forName` (which I believe is the source of the `ClassNotFoundException`) outside the container, and everyone should be happy. 😀 

Could you please test if this solves the problem? No need to update the code in the PR, I've tested and it works fine here. 

PS: It would also be nice if we could create a test case for this issue (or create a test-only module for OSGI interop in general, if it's hard to write these are normal unit tests). 


Best regards,

-- 
Harald K

koenr-bc:
Hey Harald, 

Thanks for the quick response!
That change is good, I don't think an OSGi test would be feasible (notoriously hard to setup in my opinion)

Are you incorporating that change in a separate commit? than I can close/cancel this PR.

Greetings,
Koen

haraldk:
I was just cleaning up for a 3.8.1 release, so I added it there too. Hope to release it today (if I can find time between Christmas preparations). If not, sometime soon...

koenr-bc:
Fixed via f50178bc78af3d6b7f288914a1cea121330312b8

PR #656, opened 2022-01-12
Support writing ASCII array in TIFF metadata

Schmidor:
Currently the TIFF reader supports reading fields which contain multiple ASCII strings. Writing those metadata again then fails, because writing ASCII string arrays is not supported.

### DIFF
--- a/imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFWriter.java
+++ b/imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFWriter.java
@@ -292,15 +292,17 @@ public final class TIFFWriter extends MetadataWriter {
 
     private int getCount(final Entry entry) {
         Object value = entry.getValue();
-        if(value instanceof String){
+        if (value instanceof String) {
             return ((String) value).getBytes(StandardCharsets.UTF_8).length + 1;
-        } else if(value instanceof String[]){
+        }
+        else if (value instanceof String[]) {
             int sum = 0;
-            for (String string : (String[]) value){
+            for (String string : (String[]) value) {
                 sum += string.getBytes(StandardCharsets.UTF_8).length + 1;
             }
             return sum;
-        } else {
+        }
+        else {
             return entry.valueCount();
         }
     }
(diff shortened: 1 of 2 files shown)


imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFWriter.java
```diff
@@ -410,7 +420,14 @@ else if (value instanceof long[]) {
 
                         break;
                     }
-
+                case TIFF.TYPE_ASCII:
+                    String[] strings = (String[]) value;
+                    for (String string : strings) {
+                        byte[] bytes = string.getBytes(StandardCharsets.UTF_8);
+                        stream.write(bytes);
+                        stream.write(0);
```
haraldk, line 428: Perhaps extract this (and the code below for single string) to a `writeString(stream, string)` method?

imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFWriter.java
```diff
@@ -292,7 +292,17 @@ private long writeValue(final Entry entry, final long dataOffset, final ImageOut
...
+            return ((String) value).getBytes(StandardCharsets.UTF_8).length + 1;
+        } else if(value instanceof String[]){
+            int sum = 0;
+            for (String string : (String[]) value){
+                sum += string.getBytes(StandardCharsets.UTF_8).length + 1;
+            }
+            return sum;
+        } else {
+            return entry.valueCount();
+        }
```
haraldk, line 305: This block has some code style issues (space after if, else on new line, space before opening brace etc).

haraldk (changes requested):
Thanks! 👍🏻 

I wasn't aware of this multiple string thing when I originally wrote it, thus the issue here... Perhaps a more efficient way would be to convert the strings to byte array right away? But I think this is fine for now.  

I had some minor comments, will merge if you address those! 😀

imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/tiff/TIFFWriter.java
Schmidor: i've extracted the duplicate code. But i'm not sure what could be changed for a more efficient converting to bytes?

haraldk: I was thinking about perhaps storing the `byte[]` instead of `String/String[]` in the `TIFFEntry`. Instead of converting the Strings to bytes twice, once for the value count, and once for the actual write...

But I don't think this is used enough to warrant that refactoring just now. I'm fine with the current solution.

haraldk (approved):

PR #657, opened 2022-01-14
findCompressionType always uses RLE if leading EOL is missing

Schmidor:
Update of the last read byte is missing since the last update. So if only the first EOL is missing, further EOLs after the lines are not detected.

It seems I lost the line in the refactoring steps when working on #626 and we had no tests checking for missing leading EOL.


haraldk:
👍🏻

PR #667, opened 2022-02-24
TIFFImageMetadata: ImageOrientation in mergeTree

Schmidor:


imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageMetadata.java
```diff
@@ -1024,6 +1024,34 @@ else if ("HorizontalPixelSize".equals(nodeName)) {
...
+                    case "rotate270":
+                        orientation = TIFFExtension.ORIENTATION_RIGHTTOP;
+                        break;
+                    case "flipvrotate90":
+                        orientation = TIFFExtension.ORIENTATION_RIGHTBOT;
+                        break;
+                    case "rotate90":
+                        orientation = TIFFExtension.ORIENTATION_LEFTBOT;
+                        break;
+                }
```
haraldk, line 1053: Can you extract this to a method (and use `return` in each `case`), I think that will make this more compact and easier to read.

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageMetadata.java
```diff
@@ -1024,6 +1024,34 @@ else if ("HorizontalPixelSize".equals(nodeName)) {
             else if ("VerticalPixelSize".equals(nodeName)) {
                 yRes = Float.parseFloat(getAttribute(child, "value"));
             }
+            else if ("ImageOrientation".equals(nodeName)) {
+                switch (getAttribute(child, "value").toLowerCase()) {
```
haraldk, line 1028: Can you also make this `null`-safe, in the odd case of malformed metatada, just in case?

haraldk (approved):
I have some comments (inline), otherwise this is a good addition! 👍🏻

imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageMetadata.java
Schmidor: Done, extracted and null check :)

haraldk (approved):

PR #722, opened 2022-12-31
Support org.w3c.dom.Document input to SVGImageReader

stanio:
_Use case:_ Allow reading an SVG source once, and tweaking its content in memory like changing viewBox, colors, and styles to produce multiple image variants in succession.

_Similar to:_ [DynamicSvgOffscreen](https://cwiki.apache.org/confluence/display/XMLGRAPHICSBATIK/DynamicSvgOffscreen) - Rendering a dynamic SVG document to an offscreen buffer (Batik HowTos)

### DIFF
--- a/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
+++ b/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
@@ -115,6 +115,10 @@ public class SVGImageReader extends ImageReaderBase {
             TranscoderInput input = new TranscoderInput(IIOUtil.createStreamAdapter(imageInput));
             rasterizer.setInput(input);
         }
+        else if (pInput instanceof Document) {
+            Document doc = (Document) pInput;
+            rasterizer.setInput(new TranscoderInput(doc));
+        }
     }
 
     public BufferedImage read(int pIndex, ImageReadParam pParam) throws IOException {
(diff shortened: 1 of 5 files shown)


imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReader.java
```diff
@@ -115,6 +118,24 @@ public void setInput(Object pInput, boolean seekForwardOnly, boolean ignoreMetad
...
+                processWarningOccurred("Document input has no root element");
+            }
+            else if (root.getLocalName() == null) {
+                processWarningOccurred("Not a namespace-aware Document input: " + root.getTagName());
+            }
+            else if (!root.getLocalName().equals("svg")
+                    || !root.getNamespaceURI().equals(SVG_NS_URI)) {
+                processWarningOccurred("Not a SVG Document input: " + root.getTagName()
+                        + " / {" + root.getNamespaceURI() + "}" + root.getLocalName());
+            }
```
stanio, line 134: Not sure if this sanity check should go here or in the `SVGImageReaderSpi.canDecodeInput()`.  My rationale is the Batik library might support more than `{http://www.w3.org/2000/svg}svg` documents, hence this is just a warning.  In any case, it would signal unsupported types during the image reading/rasterizing.

stanio:
Just noticed there's a previous #452.  It is slightly different from my suggestion here – it appears to perform some unnecessary hoops for `org.w3c.dom.svg.SVGDocument` vs. `org.w3c.dom.Document`, serializing the latter so it would be parsed back.  As far as I see `org.apache.batik.transcoder.TranscoderInput` could be initialized with any `org.w3c.dom.Document`, and given the instance is namespace-aware it is processed just fine by Batik (as proven by the test case I've added here).

haraldk:
Thanks!

I'll look into the changes as soon as I find the time!

PR #723, opened 2023-01-02
More universal SVG stream detection

stanio:
This is a suggestion for implementing universal SVG stream detection using a SAX parser that (setup will most likely be used for the final parsing, and):

-   Would handle any supported text encoding including UTF-16 and possible BOM

    Parser following [_F.1 Detection Without External Encoding Information_](https://www.w3.org/TR/xml/#sec-guessing) would be able to handle EBCDIC code pages as well;

-   Handles possible namespace prefixes and matches the SVG namespace precisely.

I haven't verified explicitly but the parsing is stopped at most at the root document element, and the performance should not be noticeably slower.  As previously, uses possible `DOCTYPE` declaration as a speculative optimization:

```xml
<!DOCTYPE svg ...>
<!--
  - Lenghty comment block...
  -->
<svg ...>
```

---

Notable differences with the existing implementation include:

-   No longer accepted:

    ```xml
    <!--
      -- Malformed
      -->
    <svg xmlns="http://www.w3.org/2000/svg">
    ```

-   No longer accepted:

    ```xml
    <svg xmlns="http://www.w3.org/2023/fake">
    ```

-   Now is accepted:

    ```xml
    <!DOCTYPE ns:svg>
    <ns:svg xmlns:ns="http://www.w3.org/2000/svg">
    ```

The following is wrongfully accepted as before:

```xml
<!DOCTYPE svg>
<svg xmlns="http://www.w3.org/2023/fake">
```

I don't think it is necessary but the last one could be made more reliable by considering the `DOCTYPE` declaration only when the name doesn't match `svg` and doesn't end with `:svg` (definitely not an SVG), or when the name matches `svg` or ends with `:svg` and the public identifier matches `-//W3C//DTD SVG 1.0//EN` or `-//W3C//DTD SVG 1.1//EN`.  If the `DOCTYPE` name is `svg` or ends with `:svg`, but the public identifier is unknown it could always fall back to looking at the root document element.

### DIFF
--- a/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReaderSpi.java
+++ b/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReaderSpi.java
@@ -34,21 +34,13 @@ import com.twelvemonkeys.imageio.spi.ImageReaderSpiBase;
 import com.twelvemonkeys.imageio.util.IIOUtil;
 import com.twelvemonkeys.lang.SystemUtil;
 
-import org.xml.sax.Attributes;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
-import org.xml.sax.XMLReader;
-import org.xml.sax.ext.DefaultHandler2;
-import javax.xml.XMLConstants;
-import javax.xml.parsers.FactoryConfigurationError;
-import javax.xml.parsers.ParserConfigurationException;
-import javax.xml.parsers.SAXParserFactory;
 import javax.imageio.ImageReader;
 import javax.imageio.spi.ServiceRegistry;
 import javax.imageio.stream.ImageInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.StringReader;
 import java.util.Locale;
 
 import static com.twelvemonkeys.imageio.util.IIOUtil.deregisterProvider;
@@ -77,23 +69,15 @@ public final class SVGImageReaderSpi extends ImageReaderSpiBase {
         return pSource instanceof ImageInputStream && canDecode((ImageInputStream) pSource);
     }
 
-    @SuppressWarnings("StatementWithEmptyBody")
     private static boolean canDecode(final ImageInputStream pInput) throws IOException {
-        DoctypeHandler doctype = new DoctypeHandler();
+        DoctypeHandler doctype;
         pInput.mark();
         try {
-            XMLReader xmlReader = XMLReaderFactory.getXMLReader();
-            xmlReader.setContentHandler(doctype);
-            xmlReader.setErrorHandler(doctype);
-            xmlReader.setEntityResolver(doctype);
             @SuppressWarnings("resource")
             InputStream stream = IIOUtil.createStreamAdapter(pInput);
             // XMLReader.parse() generally closes the input streams but the
             // stream adapter prevents closing the underlying stream (✔)
-            xmlReader.parse(new InputSource(stream));
-        }
-        catch (StopParseException e) {
-            // Found root element
+            doctype = DoctypeHandler.ofSource(new InputSource(stream));
         }
         catch (SAXException e) {
             // Malformed XML, or not an XML at all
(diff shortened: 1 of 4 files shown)


stanio:
>   I haven't verified explicitly but the performance should not be noticeably slower.

I've now run a few microbenchmarks.  I've measured ~54% slower performance.  I've got average times of ~68μs (current) vs. ~105μs (new).  This roughly translates into ~3.7s added time for scanning 100,000 files.  Arguably the more reliable detection is worth the possibly unnoticeable performance hit.

haraldk:
Thank you!

I'll review the changes as soon as I find the time. 

I'm initially a little reluctant to merge these changes, however, as file format detection should prioritize performance, over 100% reliable detection. From the API docs of `canDecode(Object)`: *"Returning true from this method does not guarantee that reading will succeed, only that there appears to be a reasonable chance of success based on a brief inspection of the stream contents"*. There are also some security considerations that need to be assessed, regarding XML parsing.

I also believe that [comments are allowed anywhere after the optional XML declaration](https://www.w3.org/TR/xml/#NT-prolog), so your first "no longer accepted" example above, should still be accepted. 😀

stanio:
>   I also believe that [comments are allowed anywhere after the optional XML declaration](https://www.w3.org/TR/xml/#NT-prolog), so your first "no longer accepted" example above, should still be accepted. 😀

Note, that example uses a malformed [comment](https://www.w3.org/TR/xml/#sec-comments) – a [fatal error](https://www.w3.org/TR/xml/#dt-fatal) after which XML parsers generally stop further processing:

>   For compatibility, the string " `--` " (double-hyphen) MUST NOT occur within comments.

It is not something I've implemented explicitly but it is driven by the XML parser implementation.

stanio:
>   file format detection should prioritize performance, over 100% reliable detection. From the API docs...

I'm unsure of the point you're making here.  If the format detection rejects a valid input it won't get to be decoded – that's my concern.  I don't insist on using a parser if this could be achieved better otherwise, but it seems the custom parsing in `canDecode` is already rather complex.

stanio:
> There are also some security considerations that need to be assessed, regarding XML parsing.

You may have a look at:

https://github.com/haraldk/TwelveMonkeys/blob/bc31399c6e9134b52bd5472a0984f17b987b833c/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/DoctypeHandler.java#L138-L140

and:

https://github.com/haraldk/TwelveMonkeys/blob/bc31399c6e9134b52bd5472a0984f17b987b833c/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/DoctypeHandler.java#L127-L132

Basically, no external entities are allowed – that's also necessary for performance.

haraldk:
Thanks!

You are right about the XML comment, I missed the illegal double hyphens inside the comment. 

Performance is critical, because it impacts not only the detection of SVG, but also all other formats when the SVG (Batik) plugin is installed. Most file formats can be detected just by looking at the first 2-4 bytes, without any extra allocations or instantiations. It's possible to do some tweaks to make this impact smaller, by "moving" the SVG detection to the end, after other formats. And, yes, the `ThreadLocal` probably helps a bit, but it also comes with cost and complexity... 

`ImageReaderSpi.canDecodeInput(..)` is used from `ImageIO.getImageReaders(..)` and all the `ImageIO.read(...)` methods. It's still possible to use a separate detection method, and use `ImageIO.getImageReadersByFormatName("SVG")` for example (a common use case is some formats that can't be detected by content, but can still be identified by "external" properties, like file name, using `ImageIO.getImageReadersBySuffix(suffixOf(fileName))`). 

But I agree that the best would be if all SVGs were correctly detected by the `canDecodeInput(..)` method, so I'm still considering it. 😀

imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReaderSpi.java
```diff
@@ -64,97 +71,25 @@ public boolean canDecodeInput(final Object pSource) throws IOException {
...
-            // Need to figure out pretty fast if this is XML or not
-            int b;
-            while (Character.isWhitespace((char) (b = pInput.read()))) {
-                // Skip over leading WS
-            }
-
-            // If it's not a tag, this can't be valid XML
-            if (b != '<') {
-                return false;
-            }
```
haraldk, line 85: What if we keep this, just to very quickly filter out any file that is *not* XML, before creating stream wrapper and attempting to parse?

We might have to strengthen the whitespace skipping to handle BOMs and other multibyte whitespace chars. But it would at least make me happier, knowing we won't spend time trying to XML-parse binary files... 😀

stanio: The current detection is good enough, to be honest.  What it lacks most is likely handling UTF-16 and UTF-BOM.  Given nowadays almost everything uses just UTF-8 w/o BOM, one may consider my suggestion here superfluous. 😄  It just happens I've encountered this code and decided to give it a try.

I don't think you should spend more time looking into this PR.  Thank you for providing feedback to me also.

stanio: Alternatively, the detection using an XML parser could be included and triggered by a system property, depending on one's needs.

stanio: > ... over the weekend.

Over already.

stanio:
@haraldk, I'm not after this anymore.  Please close at your discretion.

haraldk:
Sorry for taking so long! 😬 

But I guess there isn't much demand to warrant the extra time/complexity. We can always reopen and add at a later time if the feature becomes more important. 

Anyway, thanks a lot for your effort here! 👍🏻

stanio:
At the time, I was looking at using TwelveMonkeys for a tool that creates Windows mouse cursors from SVG images, but since then I've developed it as an independent solution.

PR #725, opened 2023-01-06
JPG File fails to process

KoenDG:
I ran into this image that couldn't get processed. I've added it to this branch and to the testsuite. In case someone is interested in analyzing it.

It simply fails with:
```
[INFO] Running com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReaderTest
WARNING: Reader does not throw exception with non-declared destination: BufferedImage@3b366632: type = 8 DirectColorModel: rmask=f800 gmask=7e0 bmask=1f amask=0 ShortInterleavedRaster: width = 50 height = 50 #numDataElements 1
WARNING: Reader does not throw exception with non-declared destination: BufferedImage@70925b45: type = 9 DirectColorModel: rmask=7c00 gmask=3e0 bmask=1f amask=0 ShortInterleavedRaster: width = 50 height = 50 #numDataElements 1
[ERROR] Tests run: 141, Failures: 1, Errors: 0, Skipped: 3, Time elapsed: 20.693 s <<< FAILURE! - in com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReaderTest
[ERROR] com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReaderTest.testSetDestinationType  Time elapsed: 0.71 s  <<< FAILURE!
java.lang.AssertionError: Could not read file:...TwelveMonkeys/imageio/imageio-jpeg/target/test-classes/test_jpeg/Portugal-Nazare.jpg with explicit destination type javax.imageio.ImageTypeSpecifier@ab7ef9a8
Caused by: javax.imageio.IIOException: Destination type from ImageReadParam does not match!
```

It says "could not read file", though it's perfectly accessible. The "with explicit destination" is something I don't know about.

### DIFF
--- a/imageio/imageio-jpeg/src/test/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGImageReaderTest.java
+++ b/imageio/imageio-jpeg/src/test/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGImageReaderTest.java
@@ -94,6 +94,7 @@ public class JPEGImageReaderTest extends ImageReaderAbstractTest<JPEGImageReader
     protected List<TestData> getTestData() {
         // While a lot of these files don't conform to any spec (Exif/JFIF), we will read these.
         return Arrays.asList(
+                new TestData(getClassLoaderResource("/test_jpeg/Portugal-Nazare.jpg"), new Dimension(13622, 5070)),
                 new TestData(getClassLoaderResource("/jpeg/cmm-exception-adobe-rgb.jpg"), new Dimension(626, 76)),
                 new TestData(getClassLoaderResource("/jpeg/cmm-exception-srgb.jpg"), new Dimension(1800, 1200)),
                 new TestData(getClassLoaderResource("/jpeg/corrupted-icc-srgb.jpg"), new Dimension(1024, 685)),


haraldk:
Thanks @KoenDG!

The image seems okay, no problem processing it. 

There is only a minor problem reading it with explicit target destination (as in `ImageReadParam.setDestination(...)`), because our reader reports it also supports reading as `TYPE_INT_RGB`. But this only works if we run through the raster conversion code branch, not if we read directly through the delegate (which happens for this image)... Might have to strengthen that test a bit.

I'd rather not have this image in the repo or use in test cases due to the size of the image, sorry. 😀
If you ran into a problem processing this image in production, it's more likely that you just ran out of memory.

KoenDG:
Well, it is a big file at 14MB, but bigger files of 50MB and 100MB go through without issue. That's why I thought to bring it up.

Though it may well be down to local code. I'll verify that first, sometime next week(or later).

Sorry for the interruption.

haraldk:
File size isn't that important for memory usage. The dimensions and bit depth does, so this file will decode to about 200 MB in memory.

haraldk:
Closing this. We might want to file a separate issue for the incorrect `ImageTypeSpecifier`s returned by `getImageTypes()`.

KoenDG:
I lost sight of this, so tried it today, the image requires about 1.5GB of free heap to avoid crashing.

Looks like the exception I saw in the test is suppressed, only output I'm really getting is that the image doesn't have an Orientation tag, but that's not important, I would think.

haraldk:
In my standalone test program  the file reads fine using 300 GB of heap (`-Xmx300m`) in all Java versions I've tested. For Java 11 and 17, I needed just a bit over 200 (ie. `-Xmx205m`):

    java -cp common/common-lang/target/classes:common/common-io/target/classes:common/common-image/target/classes:imageio/imageio-core/target/classes:imageio/imageio-metadata/target/classes:imageio/imageio-jpeg/target/classes -Xmx205m com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReader ~/Downloads/Portugal-Nazare.jpg 

Output: 

```
Reading using: com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReader@67424e82
[..............................................................................]
```

>  only output I'm really getting is that the image doesn't have an Orientation tag

I'm really not sure where this output comes from... I think you need to send the code to reproduce. Perhaps open a new "trouble-shooting" issue, as this discussion isn't really that suited for a PR.

PR #728, opened 2023-01-11
wip: Support fake alpha for BMP files

### LINKED ISSUE #727: Support fake alpha in BMPImageReader
**Is your feature request related to a use case or a problem you are working on? Please describe.**
Some user-supplied BMP images use fake alpha. Meaning that even though they use RGB compression, 

**Describe the solution you'd like**
Use a heuristic to determine if a BMP is using fake alpha. In particular, if any of the unused bits are non-zero for any pixel when using 32 bpp and a compression method of RGB (no bitfields), then assume it is used as alpha.

**Describe alternatives you've considered**
Another alternative would be to unconditionally assume that if there are 32 bpp it is using fake alpha. But that isn't necessarily a good assumption, and I don't even know if it is more common for BMPs with 32 bpp to use fake alpha or not. 

** Possible implementations **

This is a little tricky because you can't know if the image is using an alpha channel until you have read the pixels. 

I can think of a few ways to do this:

1. Have `getRawImageType` do a pass over the pixels if the bitCount is 32 and `hasMasks` is false to determine if any unused bits are used
2. Have `getRawImageType` return `TYPE_INT_ARGB` if the bitCount is 32 and `hasMasks` is false, then in `read(` we keep track of whether we have encountered any pixels with a non-zero alpha. Then if all pizels have an alpha of zero, change them all to have an alpha of 255. And possibly if we encounter a pizel with an alpha of zero, but non-zero RGB, then convert all previous pixels to be opaque, and switch into a mode where we set all future pixels to be opaque. (or maybe there at the end we just create a new BufferedImage with a different colormodel that ignores the alpha?).
3. Similar to 2, but reversed. We start assuming that the extra bits aren't used for alpha, and set the alpha to fully opaque, until we find a nonzero value in the extra bits, then we change all previous pixels to have an alpha of 0, and switch into a mode where we assume the unused bits contain the alpha value.

**Additional context**
See the q/rgb32fakealpha.bmp in the [bmp suite](https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html)

tmccombs:
Fixes #727

This isn't currently in a mergeable state, because it just always assumes that the unused bits are an alpha channel if bpp is 32 and compression doesn't set BITFIELDS.

### DIFF
--- a/imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReader.java
+++ b/imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/BMPImageReader.java
@@ -251,8 +251,14 @@ public final class BMPImageReader extends ImageReaderBase {
                         );
                     }
 
+                    // TODO: use some heuristic to determine if we should treat the extra byte as an alpha channel
+                    //
+                    // According to https://entropymine.com/jason/bmpsuite/bmpsuite/html/bmpsuite.html some BMP decoders
+                    // will treat the unused bits in a 32 bit representation as an alpha channel if any of the unused bits is non-zero
+                    // but that potentially requires doing a full pass over all the pizels (in that case that it doesn't use fake alpha).
+
                     // Default if no mask
-                    return ImageTypeSpecifiers.createFromBufferedImageType(BufferedImage.TYPE_INT_RGB);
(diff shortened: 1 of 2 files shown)


KoenDG:
Was going over github and chanced upon this.

@tmccombs Is this still a thing? Reading the Issue, it seems a different solution was found. Can this and this issue be closed or is there still a reason for it?

haraldk:
Closing, as the suggested workaround in #727 seems sufficient for now.

Issue #736, opened 2023-03-14
WebP: Incorrect decoding of some images with alpha channel, if source subsampling is used


tc-wleite:
**Describe the bug**
Some WebP images with alpha channel are not decoded properly when source subsampling is used.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use:
3.9.4. Also tested with the master branch.

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 
    openjdk version "11.0.13" 2021-10-19 LTS
    OpenJDK Runtime Environment (build 11.0.13+8-LTS)
    OpenJDK 64-Bit Server VM (build 11.0.13+8-LTS, mixed mode, sharing)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. 
I tested in Windows 10 and 11. 
While reading a few thousand WebP images from different sources, types and resolutions, I found this issue. I tested later with a simple standalone program.

**To Reproduce**
Run the standalone program below, using the WebP ImageIO plugin.

**Expected behavior**
Decode image using source subsample should be similar to the image decode without source subsample (only smaller, obviously).

**Example code**
```java
import java.awt.image.BufferedImage;
import java.io.File;
import java.io.IOException;
import java.util.Iterator;

import javax.imageio.ImageIO;
import javax.imageio.ImageReadParam;
import javax.imageio.ImageReader;
import javax.imageio.stream.ImageInputStream;

public class WebPSubsampligTest {
    public static void main(String[] args) throws IOException {
        int sampling = 2; // Sampling factor
        File input = new File("A.webp"); // Input image
        File output = new File("A.png"); // Input image

        ImageInputStream iis = ImageIO.createImageInputStream(input);
        Iterator<ImageReader> iter = ImageIO.getImageReaders(iis);
        ImageReader reader = iter.next();
        reader.setInput(iis, false, true);
        int w0 = reader.getWidth(0);
        int h0 = reader.getHeight(0);
        int dstW = (int) Math.ceil((double) w0 / sampling);
        int dstH = (int) Math.ceil((double) h0 / sampling);
        ImageReadParam params = reader.getDefaultReadParam();
        BufferedImage image = reader.getImageTypes(0).next().createBufferedImage(dstW, dstH);
        params.setDestination(image);
        params.setSourceSubsampling(sampling, sampling, 0, 0);
        reader.read(0, params);
        reader.dispose();
        iis.close();
        ImageIO.write(image, "png", output);
[...]

PR #737, opened 2023-03-15
WebP: Fix alpha decoding when source subsampling is used

tc-wleite:
Analyzing the issue described in #736, I believe it is caused by the way alpha channel may be encoded.

In `readAlpha()` method (referenced below), the alpha channel is decoded into the alpha raster of `destination` , which can be smaller than the original image, if **source subsampling** is used.

https://github.com/haraldk/TwelveMonkeys/blob/29f7547a99da1719aa22e7ea1f86b6b93d2fed55/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReader.java#L513

However, in the end of this method, **filters** may be applied:
https://github.com/haraldk/TwelveMonkeys/blob/29f7547a99da1719aa22e7ea1f86b6b93d2fed55/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReader.java#L551-L558

Inside the `getPredictorAlpha()` function, the values from the previous row and/or column are used to calculate the current alpha value, as shown below.
In this case, if subsampling was already applied (as it is today), this function will get a different value (from 2 or more rows/columns before). And as this is an accumulative process, pixels to the right and bottom of the image may have higher errors (as observed in the provided sample images).

https://github.com/haraldk/TwelveMonkeys/blob/29f7547a99da1719aa22e7ea1f86b6b93d2fed55/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReader.java#L561-L589

So I think the solution would be first decoding the alpha channel in a raster with the original dimensions (including applying any filters), then subsample this alpha raster to the destination raster.

As I am not very familiar with TwelveMonkeys' code, probably this can be achieved with a better implementation.
My main goal was to confirm that the proposed approach solves the issue.


haraldk:
Thank you very much! 😀

Code looks reasonable! I'll read though it in more detail and hope to merge it soon.

haraldk:
I'll merge your PR, as it seems to fix the issue. 👍🏻 

However, I would really appreciate if you could also create a test method (in the `WebPImageReaderTest` class) that fails in the current version, but passes with your fix, to avoid future regressions for subsampled reading. Preferably using one of the existing sample files. You should find plenty of examples in the existing test classes. Let me know if you have the time to provide that, or have any questions!

tc-wleite:
> I'll merge your PR, as it seems to fix the issue. 👍🏻

That is great! Thank you!
 
> However, I would really appreciate if you could also create a test method (in the `WebPImageReaderTest` class) that fails in the current version, but passes with your fix, to avoid future regressions for subsampled reading. Preferably using one of the existing sample files. You should find plenty of examples in the existing test classes. Let me know if you have the time to provide that, or have any questions!

Sure, I can try to do that. 
I will let you know if I have any questions, or update here when I am done.

tc-wleite:
> I would really appreciate if you could also create a test method (in the `WebPImageReaderTest` class) that fails in the current version, but passes with your fix, to avoid future regressions for subsampled reading. 

I just added an unit test that passes with the new code but fails with the current version.

> Preferably using one of the existing sample files.

Unfortunately, none of the existing sample files have alpha transparency and alpha filtering, which is necessary to trigger the issue, so I had to add a new sample file. Let me know if there is another way to avoid adding it.

While running the existing unit tests and creating the new one, I found two issues in the code I submitted, fixed now by the commits https://github.com/haraldk/TwelveMonkeys/pull/737/commits/3eabc591d83f4c392c19d8dca01612413b98d939 and https://github.com/haraldk/TwelveMonkeys/pull/737/commits/0a2efb9eacfc32f126b51f93654203f6952df9c7.

imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReader.java
```diff
@@ -527,7 +527,8 @@ private void readAlpha(BufferedImage destination, ImageReadParam param, final in
             System.out.println("compression: " + compression);
         }
 
-        WritableRaster alphaRaster = destination.getAlphaRaster();
+        // Alpha raster must have same dimensions as the source, because of gradient filtering.
+        WritableRaster alphaRaster = Raster.createInterleavedRaster(DataBuffer.TYPE_BYTE, width, height, 1, null);
```
haraldk, line 531: I think we should be able to get away with *not* creating another full-size temporary raster here. Instead we can do the filtering on the `tempRaster` below, by creating a *writable* child raster on line 544, and do the alpha filtering on that.

Something like this:

```java
        WritableRaster alphaRaster = destination.getAlphaRaster();

        switch (compression) {
            case 0:
                readUncompressedAlpha(alphaRaster);
                break;

            case 1:
                // Simulate header
                imageInput.seek(imageInput.getStreamPosition() - 5);

                WritableRaster tempRaster = Raster.createInterleavedRaster(DataBuffer.TYPE_BYTE, width, height, 4, null);
                readVP8Lossless(tempRaster, null, width, height);

                // Copy from green (band 1) in temp to alpha in destination
                WritableRaster alphaChannel = tempRaster.createWritableChild(0, 0, tempRaster.getWidth(), tempRaster.getHeight(), 0, 0, new int[]{1});
                alphaFilter(alphaChannel, filtering);
                alphaRaster.setRect(alphaChannel); // TODO: Replace this with copyIntoRasterWithParams(alphaChannel, alphaRaster, param);
                break;

            default:
                processWarningOccurred("Unknown WebP alpha compression: " + compression);
                opaqueAlpha(alphaRaster);
                break;
        }

```

...and extracting the current alpha filtering to a separate method:

```java
    private void alphaFilter(WritableRaster alphaRaster, int filtering) {
        if (filtering != AlphaFiltering.NONE) {
            for (int y = 0; y < alphaRaster.getHeight(); y++) {
                for (int x = 0; x < alphaRaster.getWidth(); x++) {
                    int predictorAlpha = getPredictorAlpha(alphaRaster, filtering, y, x);
                    alphaRaster.setSample(x, y, 0, alphaRaster.getSample(x, y, 0) + predictorAlpha % 256);
                }
            }
        }
    }

```

I'm not sure if alpha filtering is supported for uncompressed alpha, and we don't support it anyway (if you have a sample file with uncompressed alpha, please send it to me! 😀 ).

imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/lossless/VP8LDecoder.java
```diff
@@ -135,33 +135,44 @@ public void readVP8Lossless(final WritableRaster raster, final boolean topLevel,
...
+        Rectangle sourceRegion = param.getSourceRegion();
+        int sourceXSubsampling = param.getSourceXSubsampling();
+        int sourceYSubsampling = param.getSourceYSubsampling();
+        int subsamplingXOffset = param.getSubsamplingXOffset();
+        int subsamplingYOffset = param.getSubsamplingYOffset();
+        Point destinationOffset = param.getDestinationOffset();
+
+        if (sourceRegion == null) {
+            sourceRegion = dstRaster.getBounds();
+        }
```
haraldk, line 160: `param` null safe fix:

```suggestion
        Rectangle sourceRegion = param != null && param.getSourceRegion() != null ? param.getSourceRegion() : dstRaster.getBounds();
        int sourceXSubsampling = param != null ? param.getSourceXSubsampling() : 1;
        int sourceYSubsampling = param != null ? param.getSourceYSubsampling() : 1;
        int subsamplingXOffset = param != null ? param.getSubsamplingXOffset() : 0;
        int subsamplingYOffset = param != null ? param.getSubsamplingYOffset() : 0;
        Point destinationOffset = param != null ? param.getDestinationOffset() : new Point(0, 0) ;
```
This should make the `param == null` case fall into the fast case using `setRect` so no big performance penalty.

haraldk (changes requested):

tc-wleite:
Thanks! 
I am taking a closer look in the comments you made, and will try to make the suggested adjustments.

imageio/imageio-webp/src/test/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReaderTest.java
```diff
@@ -58,7 +58,9 @@ protected List<TestData> getTestData() {
                         new Dimension(400, 400), new Dimension(400, 400), new Dimension(400, 394),
                         new Dimension(371, 394), new Dimension(394, 382), new Dimension(400, 388),
                         new Dimension(394, 383), new Dimension(394, 394), new Dimension(372, 394),
-                        new Dimension(400, 400), new Dimension(320, 382))
+                        new Dimension(400, 400), new Dimension(320, 382)),
+                // Alpha transparency and Alpha filtering
+                new TestData(getClassLoaderResource("/webp/alpha_filter.webp"), new Dimension(2000, 1447))
```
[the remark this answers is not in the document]
tc-wleite: > If you have a sample file with smaller dimensions, that would probably be better for test performance. If not, no worries, we can use this one. 😀

I couldn't find a small sample image, so I replaced by the smallest I found, with ~15% less pixels than before (and a much smaller file in bytes).
 
> PS: There seems to be something odd with the indentation here.

I tried to follow the existing indentation. 
It looks fine for me here...
Not sure if is that what called your attention, buy the `TestData` above the one I added has multiple dimensions, so its indentation is different, right?

imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReader.java
```diff
@@ -549,13 +550,22 @@ private void readAlpha(BufferedImage destination, ImageReadParam param, final in
...
             }
         }
+
+        // Copy into destination raster
+        WritableRaster dstRaster = destination.getAlphaRaster();
+        if (param == null) {
+            dstRaster.setRect(alphaRaster);
+        } 
+        else {
+            VP8LDecoder.copyIntoRasterWithParams(alphaRaster, dstRaster, param);
```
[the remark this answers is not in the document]
tc-wleite: Agreed! I made both changes.

tc-wleite:
Ok, I think I covered all the requested changes.
Just let me if something is missing or can be improved.

imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/WebPImageReader.java
tc-wleite: > > I'm not sure if alpha filtering is supported for uncompressed alpha, and we don't support it anyway (if you have a sample file with uncompressed alpha, please send it to me! 😀 ).
> 
> I checked all WebP files I have here, and couldn't find any with uncompressed alpha. I will try to collect new ones, and let you know if I can find any of these.

Looking at a larger set of WebPs (~40K files), I found 3 that are reported as having "alpha compression = 0".
I am not sure if they actually have uncompressed alpha, or are false detections caused by some issue in the files.
They do enter in `readUncompressedAlpha()` while decoding.
[possibly-uncompressed-alpha-samples.zip](https://github.com/haraldk/TwelveMonkeys/files/10985381/possibly-uncompressed-alpha-samples.zip)

haraldk:
Excellent! You have addressed all my concerns. 👍🏻

PR #771, opened 2023-05-28
PSD: Adding parsing for 'lsdk' (undocumented) additional layer information

daaaaa:
Photoshop currently has an additional not documented key for "additional layer information" that is `lsdk`.

This should be a `nested section divider setting` and should have the same behaviour of `lsct`, that is`section divider setting`.

Other PSD libraries parse it in an equivalent fashion:
- [webtoon/psd](https://github.com/webtoon/psd/blob/main/packages/psd/src/sections/LayerAndMaskInformation/AdditionalLayerInfo/index.ts#L76)
- [psd-tools/psd-tools](https://github.com/psd-tools/psd-tools/blob/main/src/psd_tools/api/psd_image.py#L589)

And here some some github discussions [link1](https://github.com/layervault/psd.rb/issues/38) and [link2](https://github.com/psd-tools/psd-tools/issues/16).

Unfortunately I cannot share the PSD I am working on since it's covered by an NDA and I was not able to reproduce it by myself.

### DIFF
--- a/imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSD.java
+++ b/imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSD.java
@@ -709,4 +709,6 @@ interface PSD extends com.twelvemonkeys.imageio.metadata.psd.PSD {
     int luni = 'l' << 24 | 'u' << 16 | 'n' << 8 | 'i';
     int lyid = 'l' << 24 | 'y' << 16 | 'i' << 8 | 'd';
     int lsct = 'l' << 24 | 's' << 16 | 'c' << 8 | 't';
+    // Undocumented: Nested section divider setting
+    int lsdk = 'l' << 24 | 's' << 16 | 'd' << 8 | 'k';
 }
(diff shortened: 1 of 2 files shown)


imageio/imageio-psd/src/main/java/com/twelvemonkeys/imageio/plugins/psd/PSD.java
```diff
@@ -709,4 +709,5 @@ interface PSD extends com.twelvemonkeys.imageio.metadata.psd.PSD {
     int luni = 'l' << 24 | 'u' << 16 | 'n' << 8 | 'i';
     int lyid = 'l' << 24 | 'y' << 16 | 'i' << 8 | 'd';
     int lsct = 'l' << 24 | 's' << 16 | 'c' << 8 | 't';
+    int lsdk = 'l' << 24 | 's' << 16 | 'd' << 8 | 'k';
```
haraldk, line 712: Can you add just a line comment here, like: "// Undocumented: Nested section divider setting" or some similar thing, just so we don't forget what it is here for?

haraldk (commented):
I'm usually reluctant to PRs without tests, but this seems warranted...

daaaaa:
> I'm usually reluctant to PRs without tests, but this seems warranted...

I agree 100% on that.
As soon as I will be able to create a file with that key I'll share it here.

haraldk (approved):

haraldk:
Thanks!

PR #801, opened 2023-08-22
Batik version bump

KoenDG:
Compiled locally, so version 1.17 should be available.

3 security fixes:

https://issues.apache.org/jira/browse/BATIK-1349
https://issues.apache.org/jira/browse/BATIK-1347
https://issues.apache.org/jira/browse/BATIK-1346

Something about defining a list for running javascript and then disabling the previous default which allowed everything to run, without configuration.

### DIFF
--- a/imageio/imageio-batik/pom.xml
+++ b/imageio/imageio-batik/pom.xml
@@ -17,7 +17,7 @@
 
     <properties>
         <project.jpms.module.name>com.twelvemonkeys.imageio.batik</project.jpms.module.name>
-        <batik.version>1.16</batik.version>
+        <batik.version>1.17</batik.version>
     </properties>
 
     <build>


haraldk (commented):
Unfortunately, the changes (in Batik) breaks the tests. Do you have time to look into it?

KoenDG:
Just so we're on the same page: I don't actually work on batik. I had to ask something on the mailing list once and now I get these emails when a new release comes out.

I had a look, it looks like they reversed their policy on allowing external resources. You now need to explicitly allow it.

Which can be seen here: https://github.com/apache/xmlgraphics-batik/compare/batik-1_16...1_17#diff-a09c2e890f1e06f62bf5aa58e89994fddb18d88e526a7bab80aef1d55af657d9

Apparently you need to set `-allowExternalResources` as a flag? But I don't see how.

Looks like the documentation isn't updated yet either: https://people.apache.org/~clay/batik/security.html


So yeah, no idea, sorry.

haraldk:
No worries, I think I figured out how to pass my parameter on to Batik, to enable/disable the stricter check.

haraldk:
Duplicate of #802

PR #859, opened 2023-11-05
SpotBugs bug on bitwise OR of signed byte computed

KoenDG:
In trying out more static analysis, I ran into spotbugs, which produced this.

It seems to be a very niche, specific issue, only appearing for signed bytes, but also so extremely unlikely to be detected if you have to look for it. So I thought I'd add it, since the fix seems to be safe in all cases, from what I understand.

Though the test suites will have to indicate that in the end.

The detection is described here: https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bit-bitwise-or-of-signed-byte-value-bit-ior-of-signed-byte

I'll copy the explanation here:

Loads a byte value (e.g., a value loaded from a byte array or returned by a method with return type byte) and performs a bitwise OR with that value. Byte values are sign extended to 32 bits before any bitwise operations are performed on the value. Thus, if `b[0]` contains the value `0xff`, and `x` is initially 0, then the code `((x << 8) | b[0])` will sign extend `0xff` to get `0xffffffff`, and thus give the value `0xffffffff` as the result.

In particular, the following code for packing a byte array into an int is badly wrong:
```
int result = 0;
for (int i = 0; i < 4; i++) {
    result = ((result << 8) | b[i]);
}
```
The following idiom will work instead:
```
int result = 0;
for (int i = 0; i < 4; i++) {
    result = ((result << 8) | (b[i] & 0xff));
}
```

### DIFF
--- a/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReader.java
+++ b/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReader.java
@@ -1325,12 +1325,12 @@ public final class TIFFImageReader extends ImageReaderBase {
                     // but has the correct offset to the JPEG stream in the StripOffsets tag.
                     long realJPEGOffset = jpegOffset;
 
-                    short expectedSOI = (short) (imageInput.readByte() << 8 | imageInput.readByte());
+                    short expectedSOI = (short) (imageInput.readByte() << 8 | (imageInput.readByte() & 0xff));
                     if (expectedSOI != (short) JPEG.SOI) {
                         if (stripTileOffsets != null && stripTileOffsets.length == 1) {
                             imageInput.seek(stripTileOffsets[0]);
 
-                            expectedSOI = (short) (imageInput.readByte() << 8 | imageInput.readByte());
+                            expectedSOI = (short) (imageInput.readByte() << 8 | (imageInput.readByte() & 0xff));
                             if (expectedSOI == (short) JPEG.SOI) {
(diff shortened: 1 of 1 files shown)


KoenDG:
Welp, test failures. Though with specific changed output messages. 

It seems to now be encountering this: https://github.com/haraldk/TwelveMonkeys/blob/fef8ff3aab76786ac5d14f4f88569669590e91e5/imageio/imageio-tiff/src/main/java/com/twelvemonkeys/imageio/plugins/tiff/TIFFImageReader.java#L1374

Where before it did not.

I can't tell the significance of this, I'm afraid.

haraldk:
Good catch! Interesting find. 😀 

Just because something "works" doesn't necessarily mean it's _correct_... I'll need to dig a bit further to understand why the current (incorrect) code seems to produce good results, and why the fix don't. Might be more issues with the logic. 

But I think as the JPEG markers are always `0xFFnn` (and the markers we're interested in are SOI; `0xFFD8` and SOS; `0xFFDA`), the sign extension will just "mask" the first byte which should be `0xFF` anyway...

KoenDG:
Neat. Glad to hear it's something actually worth looking in to. I just hope I'm not wasting your time with PRs like this.

haraldk:
As you might have seen, I'm doing a major TIFF reader clean-up/refactor. I fixed the issue this PR addresses in a separate branch, that I'll merge soon. It was actually a real bug (and the test that failed was incorrect). The TIFF reader is just so lenient about minor TIFF format mistakes it decoded the image correctly anyway. 😛

KoenDG:
Neat. I'll close this PR then.

Issue #775, opened 2023-06-05
TIFF: Allow reading non-normalized/non-clamped values for floating point


butlnor:
Regarding my later comment in: https://github.com/haraldk/TwelveMonkeys/issues/740#issuecomment-1491668380

The TIFF reader properly reads the floating point values, but unfortunatelly it then clamps the values into the 0..1 range, with `clamp` method in `TIFFImageReader.java`. This is not ok, because TIFFs are not just images, but can also be containers for any data, like spatial terrain elevation. 

Floating point TIFF is normally not meant to be represented directly as grayscale or RGB values. Floating point values are most often mapped into color by some min/max value normalization or some other formula. It's wrong to speculate that 0.0 is always black and 1.0 is zero.

haraldk:
Hi @butlnor,

I understand. However, the `BufferedImage` class represents images, and various operations like `drawImage` or `BufferedImageOp`s will crash/throw exceptions for unnormalized image data. There's also the case with random users of the library who just use `ImageIO.read(anyFile)` and expects everything to "just work" (my experience is that such users don't read the docs, but still complain when "it doesn't work"...). So, to support these use cases, I like to keep the clamping as default. 

However, there are some options to consider for advanced users (none of these are in the library today):

 1. An `ImageReadParam` to control reading/normalization, perhaps similar to (a generalization of?) the `ToneMapper` used for the HDR plugin. The default would still need to be clamped values for the reasons described above, but allowing other normalizations or no normalization at all. 
 2. Using `ImageReader.readRaster(int, ImageReadParam)` instead of `read(int, ImageReadParam)`. This would be analogous to the JPEG plugin, where reading as raster will give you the "raw" YCbCr values stored (not converted to RGB), or HDR plugin, giving the "raw" RGBE values as stored in the file. The TIFF plugin could then just give you the values as stored in the file. 

I think I prefer option 2. it's not a huge task to do.

butlnor:
Ok. So you  are saying the the option 2 is already working as I expect, or you should change something first? I can go with option 2.

rfreeman853:
Another vote for option 2.

haraldk:
Ok, so option 2 it is! 👍🏻 

But no, as I wrote above, none of the options are implemented in the library today. I'll happily accept both PRs and donations. 😉

bchapuis:
[...]

Issue #740, opened 2023-03-17
TIFF: Support for predictor: 3 (horizontal floating point predictor)


butlnor:
**Describe the bug**
Where reading a 32-bit float TIFF, like for example a TIFF file from public URL:

- https://copernicus-dem-90m.s3.eu-central-1.amazonaws.com/Copernicus_DSM_COG_30_N46_00_E008_00_DEM/Copernicus_DSM_COG_30_N46_00_E008_00_DEM.tif

The twelvemonkeys lib report an error:

```
Unsupported TIFF Predictor value: 3
	at com.twelvemonkeys.imageio.plugins.tiff.TIFFImageReader.createUnpredictorStream(TIFFImageReader.java:2487)
	at com.twelvemonkeys.imageio.plugins.tiff.TIFFImageReader.read(TIFFImageReader.java:1123)
	at javax.imageio.ImageReader.read(ImageReader.java:940)
```

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.

3.9.4

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 

Any java version that I tried. Regarding the stack trace, the java version is irrelevant.

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable.

Windows 10

**Example code**
`ImageIO.read()`

haraldk:
Hi!

The message `"Unsupported TIFF Predictor value: 3"` means that your file uses horizontal floating point predictor, and while this is a known, valid predictor value according to the TIFF 6.0 spec, we don't support it (it's not implemented at the moment). So, this is not a bug, but a possible new feature.

If you need this feature, perhaps you would like to contribute the code or sponsor the project to help us implement it? 😀

butlnor:
> Hi!
> 
> The message `"Unsupported TIFF Predictor value: 3"` means that your file uses horizontal floating point predictor, and while this is a known, valid predictor value according to the TIFF 6.0 spec, we don't support it (it's not implemented at the moment). So, this is not a bug, but a possible new feature.
> 

The problem is that also no other ImageReader in Java supports these files, so I am now stuck. These files are completely legal GeoTIFF files, and GDAL, QGIS program, and IrfanViewer all open it without problems.

> If you need this feature, perhaps you would like to contribute the code or sponsor the project to help us implement it? 😀

[...]

PR #881, opened 2023-12-19
Disable clamping when reading raw tiff data

bchapuis:
This PR adresses the issue described in #740 and  #775 using the second option described by @haraldk.


haraldk:
Thank you @bchapuis ! 

I will have a look at the code and review as soon as I find the time. There's a holiday coming up, so don't worry if it takes some time. 🎄

haraldk:
Thanks! 

Have looked at the code and I think it does solve the use case from #775. 👍🏻 

However, I think I would prefer if the flag could be named `normalize`. And instead of passing it to the `normalizeColor` method(s), just avoid invoking them at all in the raster reading case. I think that would be a more general solution. Would you be able to do that?

bchapuis:
Sure, I wasn't sure if the photometric interpretation could be skipped. It is much better that way. Thanks a lot for the review.

haraldk:
Excellent! 

I see the PR still only handles the (non-)normalization for the floating point case (`normalizeColor(int, int, float[])`), my idea above was to wrap all invocations (also `normalizeColor(int, int, byte[])`, `normalizeColor(int, int, short[])` and `normalizeColor(int, int, int[])` methods) in the same logic. 

I realize this isn't completely trivial due to some code duplication (that I am working to remove in a separate branch), but if you would like to try, that would be great! Otherwise I can probably fix that later.

bchapuis:
Thank you for the feedback. 

I must admit that do this as a road crossing 🐌 , having no idea of the big picture. I included the normailzeColorPlanar method in the last commit and the tests are still passing 😅

haraldk:
Looking good! 👍🏻 

We should probably add some tests for this as well, but I can always add those later. If you have time now though, I'd be very happy if you could add just one test making sure that `readRaster` does indeed return unnormalized/unclamped values for one of the floating point samples already in the test/resources/tiff folder.

bchapuis:
Yes, I also had that in mind, I struggled a bit at finding a good normalization example in the ressource directory and didn't want to add an existing DEM file as they may be subject to license requirements. I will be on holiday, but I will see if I can generate a good sample with random data when I come back.

bchapuis:
Generating a tiff containing random values was actually pretty easy with qgis. I added it to the `test/resources/tiff` directory, its size (246KB) is similar to the other files. As all the pixels are altered by the normalization, the test can iterate over all the values obtained with the `readRaster` method and assert that they are different than the pixels obtained with the `read` method. Is this sufficient?

haraldk:
Happy new year! 🎉 

Changes looks good! I would prefer if the test tested that the values were equal to the known correct value, rather than (any value) different from the incorrect value. Maybe you could test a single row against a fixed float array or so? 

Other than that, this is perfect! 😀

bchapuis:
@haraldk Great, thank you for the feedback! I modified the test with an two dimentional array of floats representing a 5 by 5 square of pixels.

haraldk (approved):
Brilliant, thank you! 😀

bchapuis:
@haraldk Thanks a lot for your work and for patiently reviewing this PR 👍

bchapuis:
@haraldk Out of curiosity, is there a timeline for the next release? I'd love to use TwelveMonkeys for reading geotiff files.

haraldk:
Sorry, no explicit release date, but it's on my list quite near the top. 😀

bchapuis:
Cool, thanks a lot for your work 😅

haraldk:
@bchapuis https://github.com/haraldk/TwelveMonkeys/releases/tag/twelvemonkeys-3.11.0

bchapuis:
@haraldk Awesome! Thanks a lot 🚀

PR #1080, opened 2025-01-16
Downport fix commit for CVE-2021-23792 to 3.3.x

henrikplate:
This is a down-port of the fix for vulnerability CVE-2021-23792 to branch 3.3.x.

The patch is identical to the fix provided for later versions with fix commit [da4efe9](https://github.com/haraldk/TwelveMonkeys/commit/da4efe98bf09e1cce91b7633cb251958a200fc80).

It has been tested through the down-ported unit test included in the same commit.

Disclaimer and background:
- It is clear that the given version is EOL, i.e. the project may not release a new version containing the patch, and the project generally advises users to migrate to later, supported versions. However, this fix could be a temporary remedy for users struggling to perform this migration (whatever the reason).
- This fix is part of [Endor Labs](https://www.endorlabs.com/)' efforts to create minimal invasive, yet viable security fixes for older OSS versions, similar to the well-known and established approach followed by various Linux distributions.
- By sharing the patch with the project community, we also hope to receive input as to whether the proposal is viable. It is clear, however, that the time of open source maintainers is limited. Hence, we appreciate any feedback given, but also understand if you cannot find the time to comment on merge requests opened for EOL versions.

### DIFF
--- a/imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/xmp/XMPReader.java
+++ b/imageio/imageio-metadata/src/main/java/com/twelvemonkeys/imageio/metadata/xmp/XMPReader.java
@@ -28,25 +28,34 @@
 
 package com.twelvemonkeys.imageio.metadata.xmp;
 
-import com.twelvemonkeys.imageio.metadata.Directory;
-import com.twelvemonkeys.imageio.metadata.Entry;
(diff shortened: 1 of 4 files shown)


haraldk:
Thanks! I'll merge the PR, but I don't have any plans to create a new 3.3.x release at this moment. As you mention, it may still be helpful for for users who need this version, as they can build their own patched version.

PR #1085, opened 2025-01-27
Removing incompatible line with AEM's OSGi implementation.

un-tal-luis-nava:
This commit is getting rid of an unavailable service in AEM.

### DIFF
--- a/imageio/pom.xml
+++ b/imageio/pom.xml
@@ -149,19 +149,19 @@
                 <version>${project.version}</version>
             </dependency>
 
-	        <dependency>
+            <dependency>
                 <groupId>${project.groupId}</groupId>
                 <artifactId>imageio-metadata</artifactId>
                 <version>${project.version}</version>
             </dependency>
 
-	        <dependency>
+            <dependency>
                 <groupId>${project.groupId}</groupId>
                 <artifactId>imageio-jpeg</artifactId>
                 <version>${project.version}</version>
             </dependency>
 
-	        <dependency>
+            <dependency>
                 <groupId>${project.groupId}</groupId>
                 <artifactId>imageio-tiff</artifactId>
                 <version>${project.version}</version>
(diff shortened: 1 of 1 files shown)


haraldk:
Hi Luis,

If removing these services makes no difference for users of other OSGi implementations, I guess that should not be a problem. We just need to ensure that somehow. My only concern is I don't use OSGi myself or know it in enough detail to understand whether or not this is the correct fix. 

Can you write some kind of interop-test that fails with the old config and verifies the new one?

We should also remove the code complete, rather than commenting it.

The OSGi support was implemented by @steinarb (using it with Apache Felix/Karaf IIRC), maybe he has some input on this matter?

steinarb:
Removing that line will remove the support for [Apache Aries SPIfly](https://aries.apache.org/documentation/modules/spi-fly.html), which is the service that starts SPI services and makes them available as OSGi services.

So this doesn't make me jump with joy.

I have never heard about [AEM's OSGi implementation](https://experienceleague.adobe.com/en/docs/experience-manager-65/content/implementing/deploying/configuring/configuring-osgi) before today (I had to google it), but if it doesn't support tRequire-Capabiliy for osgi.serviceloader.registrar it sounds incomplete to me...?

@un-tal-luis-nava What is the exact error your are trying to fix with this?  

If you can give the exact error message I may be able to suggest a fix.

Also what is your java version? Version of all build tools, etc.

un-tal-luis-nava:
Hello @haraldk and @steinarb :
Thanks for the time you spent on the review, I'll try to implement @steinarb 's suggestion of bringing SPIfly into AEM.
- side note.. AEM is based on top of [Apache Sling](https://sling.apache.org/documentation/the-sling-engine/architecture.html#osgi), my apologies for the non-accurate language.

un-tal-luis-nava:
Hello @haraldk and @steinarb 
Thanks for the time you spent on the review, I'll try to implement @steinarb 's suggestion of bringing SPIfly into AEM.
- side note.. AEM is based on top of [Apache Sling](https://sling.apache.org/documentation/the-sling-engine/architecture.html#osgi), my apologies for the non accurate language.

steinarb:
@un-tal-luis-nava Let me know if you run into problems bringing SPIfly into Sling and I will try to assist

PR #1115, opened 2025-03-24
Improve debug output

rolfhowarth:
Include hex opcode and file offset in debug output. Condense scan line messages in readOpDirectBits().

PR #1116, opened 2025-03-24
Fix rendering of images with JPEG slices

rolfhowarth:
Some files contain multiple image "slices", each compressed with QuickTime JPEG compression. These need to take the matrix into account to combine them properly. Additionally, these images have a placeholder icon, rendered with PackBitsRect, to indicate that QuickTime is needed which should be suppressed if the CompressedQuickTime operation succeeded. See for example J19.pict.


haraldk (commented):
Again, thanks! 

I think this PR has some issues though, first of all, it contains the changes from the previous PR, making this PR hard to merge without first merging the other. 

It also contains the same formatting issues as the other (4 space/8 spaces continuation/no tabs). Let me know if you would like me to share my code style settings.

rolfhowarth:
I created a new pull request (#1124) as my previous one shouldn't have had a dependency on #1115

PR #1118, opened 2025-03-25
Make sure maskSize bytes are skipped

rolfhowarth:
If the CompressedQuickTime opcode has a mask then this wasn't processed properly. See for example P30946BDC.pict.


rolfhowarth:
Ah sorry, I forgot to add the file to getTestData()

Unfortunately I don't think without the fix it crashes opening the file as a whole. You end up with a valid image that says "QuickTime and TIFF decompressor are required". With and without the fix the results are as follows. Do you have any tests that check the contents of the decoded image?

Screenshot 2025-03-25 at 16 02 35

haraldk:
I see... That makes testing slightly harder. It doesn't hurt adding it to the `getTestData` list, but it doesn't fully verify the correct image is read. 

We need an extra test method. 

There is a method `ImageReaderAbstractTest.assertImageDataEquals(..)` that we can use, probably testing only a region of the image.  Or, we can just test a few random pixels using `image.getRGB(x, y)` and verify that they are equal to some expected value, using `ImageReaderAbstractTest.assertRGBEquals(..)`. 

Here's an example from the `WebPImageReaderTest` that can probably be adapted, verify a few more pixels and we're good: 

```java
    @Test
    public void testRec601ColorConversion() throws IOException {
        WebPImageReader reader = createReader();

        try (ImageInputStream stream = ImageIO.createImageInputStream(getClassLoaderResource("/webp/blue_tile.webp"))) {
            reader.setInput(stream);

            BufferedImage image = reader.read(0, null);
            assertRGBEquals("RGB values differ, incorrect Y'CbCr -> RGB conversion", 0xFF72AED5, image.getRGB(80, 80), 1);
        }
        finally {
            reader.dispose();
        }
    }
```

haraldk (changes requested):
Need a test case, see previous comment.

haraldk (approved):

PR #1124, opened 2025-03-31
Fix rendering of images with JPEG slices

rolfhowarth:
Some files contain multiple image "slices", each compressed with QuickTime JPEG compression. These need to take the matrix into account to combine them properly. Additionally, these images have a placeholder icon, rendered with PackBitsRect, to indicate that QuickTime is needed which should be suppressed if the CompressedQuickTime operation succeeded. See for example J19.pict.


rolfhowarth:
This replaces PR #1116. Hopefully it includes all the requested review changes! I had to create a new branch and new pull request so it doesn't depend on #1115

imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
```diff
@@ -1567,10 +1570,13 @@ private void readCompressedQT(final ImageInputStream pStream) throws IOException
...
         // Matrix
-        int[] matrix = new int[9];
+        float[] matrix = new float[9];
         for (int i = 0; i < matrix.length; i++) {
-            matrix[i] = pStream.readInt();
+            int fp = i%3==2 ? 30 : 16; // u v w are 2.30 fixed point, a b c d x y are 16.16 fixed point
+            matrix[i] = pStream.readInt() / (float)(1<<fp);
         }
+        matrix[6] = matrix[6] / (float)screenImageXRatio;
+        matrix[7] = matrix[7] / (float)screenImageYRatio;
```
haraldk, line 1579: Of curiosity, do you have any sample images that has a ratio other than 1:1? 

It could be that this is correct, I just need some convincing arguments... 😉

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReaderTest.java
```diff
@@ -82,6 +82,7 @@ protected List<TestData> getTestData() {
                 new TestData(getClassLoaderResource("/pict/FC10.PCT"), new Dimension(2265, 2593)),
                 // 1000 DPI with bounding box not matching DPI
                 new TestData(getClassLoaderResource("/pict/oom.pict"), new Dimension(1713, 1263)),
+                new TestData(getClassLoaderResource("/pict/J19.pict"), new Dimension(640, 480)),
```
haraldk, line 85: This is okay. But we also need a separate test here, that verifies the contents of the decoded image. See the comment in the other PR for an example.

haraldk (changes requested):

imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
```diff
@@ -1874,7 +1899,13 @@ else if (rowBytes > 250) {
         BufferedImage img = images.get(pPixmapCount);
         if (img != null) {
             srcRect.setLocation(0, 0); // Raster always start at 0,0
-            context.copyBits(img, srcRect, dstRect, transferMode, region);
+            if (dstRect.equals(lastQTRect)) {
```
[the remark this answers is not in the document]
rolfhowarth: Ah ok, sorry I must have missed your comment. 
Like you, I'm not 100% convinced about the logic, and I don't have any references that back up my approach, other than empirically it works. I guess my reasoning would be rather than just blindly skipping the next opcode, whatever it might be, I only suppress the copyBits opcode if it EXACTLY matches the rectangle that was just successfully rendered. I can't see why you would want to overwrite what you just rendered except as a fallback if the compressedQT render failed.
Maybe only keep lastQTRect for one subsequent opcode and then null it out? And add some extra comments explaining the rationale? But I don't know.

haraldk (approved):

PR #1129, opened 2025-04-10
Fix bounds issue in some v2 PICT files

rolfhowarth:
x2 and y2 were swapped over, resulting in an "invalid bounds" exception with -ve height on some nonV2Ext images, see for example CatDV==2.0=1=.pict or Picture14.pict


imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
```diff
@@ -251,8 +251,8 @@ else if (version == PICT.OP_VERSION && pStream.readShort() == PICT.OP_VERSION_2)
                 // the default Mac screen resolution and the image resolution
 
                 // int y, x, w(?), h (fixed point)
-                double y2 = PICTUtil.readFixedPoint(pStream);
                 double x2 = PICTUtil.readFixedPoint(pStream);
+                double y2 = PICTUtil.readFixedPoint(pStream);
                 double w2 = PICTUtil.readFixedPoint(pStream); // ?!
                 double h2 = PICTUtil.readFixedPoint(pStream);
```
haraldk, line 257: Can you also fix the comment above to say x, y, w, h order? 😀

This is what I would expect, not really sure why I thought it was y, x, h, w (thus the "?" comment on the width), but at least I knew this was dodgy...

haraldk (commented):
Looks pretty good to me! Will try to merge when I find some spare time.

imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/PICTImageReader.java
rolfhowarth: I noticed that in a few places, you seem to read variables called x y w h when I think it's actually x0 y0 x1 y1. But I didn't want to rename all the variables just for the sake of it

haraldk (approved):

PR #1131, opened 2025-04-11
Use signed arithmetic when reading rectangle

rolfhowarth:
An operation such as DirectBitsRect can fail if the origin is -ve and we're applying a screen image ratio. See for example P564B1400.pict


haraldk (changes requested):
Thanks for another PR! I'll review in more depth and hopefully merge soon! 

Hope you can have a look at the failing test! 😀

haraldk (approved):

PR #1183, opened 2025-09-10
Handle lossless JPEG with non-zero point transform

### LINKED ISSUE #1179: Non-zero point transform value not handled for lossless JPEG decompression
**Describe the bug**

Lossless JPEG can make use of a point transform parameter (Pt), defined in section A.4 of ISO/IEC 10918-1. The parameter is stored in the last four bits of the JPEG Scan header, which is the same position of the successive approximation bit position, low (Al) parameter in lossy mode. The specification indicates that when this value is non-zero, outputs of the decoder should be multiplied by 2^Pt.  Additionally, the prediction value at the beginning of the first lines and the beginning of each restart interval is specified to be 2^(P – Pt – 1), where P is the sample precision. The net effect is to reduce precision in sample quantization, which is a lossy operation that seems at odds with the goal of lossless image compression. However, the fact that it is present in the standard naturally means that some people have decided to make use of it. Currently, TwelveMonkey's lossless JPEG decoder ignores the Pt value and as a result will output incorrect values when it is non-zero.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.

3.12.0

2. The *exact* output of `java --version` (or `java -version` for older Java releases).

> openjdk 11.0.28 2025-07-15 LTS
> OpenJDK Runtime Environment Zulu11.82+19-CA (build 11.0.28+6-LTS)
> OpenJDK 64-Bit Server VM Zulu11.82+19-CA (build 11.0.28+6-LTS, mixed mode)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable.

Applicable to any use of the library to decode lossless JPEG with non-zero Pt parameter.

**To Reproduce**
I created a 256x256 test image where the upper 8 bits of each pixel value are the Y coordinate and the lower 8 bits are the X coordinate, then compressed it with lossless JPEG using a point transform value of 4. It’s attached here as gradient_ls_pt.jpg. I also have attached a version that doesn’t use the point transform parameter. Both were compressed using the lossless JPEG encoder in DCMTK’s dcmcjpeg utility. 

1. Add the below JUnit code to com.twelvemonkeys.imageio.plugins.jpeg.JPEGImageReaderTest
2. Download the attached gradient_ls_pt.jpg file and add it to the imageio/imageio-jpeg/src/test/resources/jpeg-lossless/ directory
3. Run the testReadLosslessJPEGGradientWithPointTransform test with JUnit

**Expected behavior**
The decompressed image should have pixel values that match the original input with reduced precision.

**Example code**
```
@Test
   public void testReadLosslessJPEGGradientWithPointTransform() throws IOException {
   	 JPEGImageReader reader = createReader();
        try (ImageInputStream stream = ImageIO.createImageInputStream(getClassLoaderResource("/jpeg-lossless/gradient_ls_pt.jpg"))) {
            reader.setInput(stream);
            assertEquals(256, reader.getWidth(0));
            assertEquals(256, reader.getHeight(0));
            BufferedImage image = reader.read(0, null);
            assertNotNull(image);
            assertEquals(256, image.getWidth());
            assertEquals(256, image.getHeight());
            assertEquals(16, image.getColorModel().getComponentSize(0));
           
            for(int y = 0; y < 256; y++) {
           	 for(int x = 0; x < 256; x++) {
           		 int expected = (y << 8) | x;
           		
           		 // Simulate effect of precision loss due to point transform = 4
           		 expected = (expected >> 4) << 4;
           		
           		 assertEquals(expected, image.getRaster().getSample(x, y, 0));
           	 }
            }
        }
        finally {
            reader.dispose();
        }
   }
```

**Sample file(s)**
gradient_ls.jpg : test image compressed as lossless JPEG with Pt = 0
Image

gradient_ls_pt.jpg : test image compressed as lossless JPEG with Pt = 4
Image

zcronix:
Fixes #1179


imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGLosslessDecoder.java
```diff
@@ -229,9 +231,14 @@ int[][] decode() throws IOException {
                     current = decode(pred, temp, index);
 
                     while ((current == 0) && ((xLoc < xDim) && (yLoc < yDim))) {
-                        output(pred);
+                        output(pred);
```
haraldk, line 234: This diff looks a bit unnecessary...

imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGLosslessDecoder.java
```diff
@@ -229,9 +231,14 @@ int[][] decode() throws IOException {
                     current = decode(pred, temp, index);
 
                     while ((current == 0) && ((xLoc < xDim) && (yLoc < yDim))) {
-                        output(pred);
+                        output(pred);                        
                         current = decode(pred, temp, index);
                     }
+                    
+                    if((current == JPEG.EOI) && (xLoc == xDim-1) && (yLoc == yDim-1)) {
```
haraldk, line 238: Code style issue:

```suggestion
                    if ((current == JPEG.EOI) && (xLoc == xDim-1) && (yLoc == yDim-1)) {
```

imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGLosslessDecoder.java
```diff
@@ -268,6 +275,17 @@ int[][] decode() throws IOException {
         // TODO oe: 05.05.2018 Is it correct loop? Content of outputData from previous iteration is always lost.
         } while ((current != JPEG.EOI) && ((xLoc < xDim) && (yLoc < yDim)) && (scanNum == 0));
 
+        // Apply point transform to output. This must be done after it has finished being
+        // used for predictive purposes.
+        if(scan.approxLow != 0) {
```
haraldk, line 280: Code style: 

```suggestion
        if (scan.approxLow != 0) {
```

imageio/imageio-jpeg/src/main/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGLosslessDecoder.java
```diff
@@ -268,6 +275,17 @@ int[][] decode() throws IOException {
...
+        	for (int componentIndex = 0; componentIndex < numComp; ++componentIndex) {
+        		int[] comp = outputData[componentIndex];
+        		for(int i = 0; i < comp.length; i++) {
```
haraldk, line 283: Code style: 

```suggestion
        		for (int i = 0; i < comp.length; i++) {
```

imageio/imageio-jpeg/src/test/java/com/twelvemonkeys/imageio/plugins/jpeg/JPEGImageReaderTest.java
```diff
@@ -2033,4 +2033,65 @@ public void testInfiniteLoopCorruptRaster() throws IOException {
...
+
+             BufferedImage image = reader.read(0, null);
+
+             assertNotNull(image);
+             assertEquals(256, image.getWidth());
+             assertEquals(256, image.getHeight());
+             assertEquals(16, image.getColorModel().getComponentSize(0));
+             
+             for(int y = 0; y < 256; y++) {
+            	 for(int x = 0; x < 256; x++) {
```
haraldk, line 2083:
```suggestion
             for (int y = 0; y < 256; y++) {
            	 for (int x = 0; x < 256; x++) {
```

Issue #1168, opened 2025-07-28
WebP: Reader produces color distortion (vertical green/purple lines) when using source region without subsampling


viktor-merkel:
**Describe the bug**
When reading WebP images using ImageReadParam.setSourceRegion() without subsampling, the resulting image exhibits color channel distortion appearing as vertical green and purple lines throughout the cropped region. The issue is resolved when subsampling is set to 2 or higher.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use: 3.12.0

2. The *exact* output of `java --version` (or `java -version` for older Java releases):
openjdk 11.0.14.1 2022-02-08 LTS
OpenJDK Runtime Environment Corretto-11.0.14.10.1 (build 11.0.14.1+10-LTS)
OpenJDK 64-Bit Server VM Corretto-11.0.14.10.1 (build 11.0.14.1+10-LTS, mixed mode)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable:
Microsoft Windows 10 Enterprise
Version 10.0.19045 Build 19045
Processor 11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz, 1498 MHz, 4 cores

**To Reproduce**

1. Load a WebP image using TwelveMonkeys ImageIO WebP reader
2. Set a source region using ImageReadParam.setSourceRegion()
3. Do NOT set subsampling (or set subsampling to 1)
4. Read the image
5. Display or save the resulting BufferedImage

**Expected behavior**
The image should be correctly cropped to the specified source region with proper color reproduction.

**Example code**
A standalone program with a `main` method that showcases the problem is attached. 

**Sample file(s)**
The sample webp file is inside the project.

**Stak trace**
No stack trace

**Screenshots**

<img width="576" height="476" alt="Image" src="https://github.com/user-attachments/assets/e5eb0668-5bf3-4fdd-aabc-1870dc165513" />

[SimpleWebPViewer.zip](https://github.com/user-attachments/files/21470837/SimpleWebPViewer.zip)

PR #1220, opened 2025-11-17
WebP: Fix decoding using source region without subsampling

don-vip:
Fixes the two problems that can be seen in #1168:

1. The direct copy using setRect when subSampling equals to 1 requires that source and target rasters are compatible, which isn't the case (it is always TYPE_BYTE while the provided test case uses TYPE_INT_RGB). To address this problem I propose a new method `createCompatibleRaster`
2. The coordinates are wrong in this case as sourceRegion is completely ignored. To address this problem I propose to use the `Raster.createChild` method

Result with provided test case:

image


don-vip:
Test ready, waiting for github to [fix their stuff.](https://www.githubstatus.com/incidents/5q7nmlxz30sk)..

don-vip:
@haraldk : I have pushed the test, it works.

haraldk (approved):

PR #1230, opened 2025-12-10
DDS DXT10 support, with some certain supported DXGI Formats only.

KhanhPham05:
# Changes
Changed DDS 4-bytes magic number read method from array `{'D', 'D', 'S', ' '};` to int `0x20534444`;

# Additions
- `enum DX10DXGIFormat` which specifies a certain *common* compression methods the library is supported. [(quick link)](https://github.com/KhanhPham05/TwelveMonkeysDDSDXT10/blob/master/imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DX10DXGIFormat.java)
- `class DX10Header` to accomodate data for the extra 20 bytes the DXT10 type provides as header extention.

# Description
- ~~Some restriction may applies on the compression format, which 4 bytes 0x80 - 0x83 of the file (dxgiformat) must be `DXGI_FORMAT_*_UNORM_SRGB` format (with `*` being one of the enum element), see the enum mentioned above~~
- `resourceDimention` in bytes 0x84 - 0x87 must be `D3D10_RESOURCE_DIMENSION_TEXTURE2D` which is `3`.
- More dxgiformat supports maybe added in the future, this changes should cover commonly used compression types for now.

### DIFF
--- a/imageio/imageio-dds/src/test/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageReaderTest.java
+++ b/imageio/imageio-dds/src/test/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageReaderTest.java
@@ -33,7 +33,7 @@ package com.twelvemonkeys.imageio.plugins.dds;
 import com.twelvemonkeys.imageio.util.ImageReaderAbstractTest;
 
 import javax.imageio.spi.ImageReaderSpi;
-import java.awt.*;
+import java.awt.Dimension;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;


imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DX10DXGIFormat.java
```diff
@@ -0,0 +1,44 @@
...
+
+    DX10DXGIFormat(DDSType ddsType, int... supportedFormats) {
+        this.ddsType = ddsType;
+        this.dxgiFormat = value -> Arrays.stream(supportedFormats).anyMatch(supportedFormat -> supportedFormat == value);
+    }
+
+    DX10DXGIFormat(DDSType ddsType, int dxgiFormatMin, int dxgiFormatMax) {
+        this.dxgiFormat = value -> dxgiFormatMin <= value && value <= dxgiFormatMax;
+        this.ddsType = ddsType;
+    }
```
haraldk: This code is a bit error prone, in that the first constructor would handle the cases with 0-1 and 3 or above arguments, while the second would handle the case with exactly 2 arguments... 

Maybe a better solution would be to have a single constructor with the predicate, and static methods for creating the predicates, either `exactly(int...)` or `range(int from, int to)`?

imageio/imageio-dds/src/test/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageReaderTest.java
```diff
@@ -57,6 +57,7 @@ protected List<TestData> getTestData() {
         Dimension dim1 = new Dimension(1, 1);
 
         return Arrays.asList(
+                new TestData(getClassLoaderResource("/dds/dxt10.dds"), dim256),
```
haraldk: Excellent, thanks for providing a test case! I prefer, however, if you append it to the end of the list. 😀

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSReader.java
```diff
@@ -83,6 +84,10 @@ int[] read(ImageInputStream imageInput, int imageIndex) throws IOException {
 
         // type
         DDSType type = getType();
+        if (type == DDSType.DXT10) {
+            this.dxt10Header = DX10Header.read(imageInput);
+            type = this.dxt10Header.getDDSType();
+        }
```
haraldk: Don't think we need the `this` here.

```suggestion
        if (type == DDSType.DXT10) {
            dxt10Header = DX10Header.read(imageInput);
            type = dxt10Header.getDDSType();
        }
```

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DX10Header.java
```diff
@@ -0,0 +1,29 @@
...
+//https://learn.microsoft.com/en-us/windows/win32/direct3ddds/dds-header-dxt10
+public final class DX10Header {
+    final DX10DXGIFormat dxgiFormat;
+    final int resourceDimension, miscFlag, arraySize, miscFlags2;
+
+    private DX10Header(int dxgiFormat, int resourceDimension, int miscFlag, int arraySize, int miscFlags2) {
+        this.dxgiFormat = DX10DXGIFormat.getFormat(dxgiFormat);
+        this.resourceDimension = resourceDimension;
+        //only support D3D10_RESOURCE_DIMENSION_TEXTURE2D = 3
+        if (this.resourceDimension != 3) throw new UnsupportedOperationException("Resource dimension " + resourceDimension + " is not supported");
```
haraldk: I think it would be better if we add a constant for `D3D10_RESOURCE_DIMENSION_TEXTURE2D` either in the `DDS` interface,  or just local this class.

KhanhPham05:
I've been thinking about a DDS writer for the already supported compression types in DDSReader.java, pretty much a reverse engineered approach based on the existing reader/decoder codes. But DDS has a lot of saving options that I don't really sure how should we do to let users to pass in those options for the to-be encoder.

Issue #978, opened 2024-07-10
Webp: "Coordinate out of bounds!" exception thrown when loading a certain file


Parsee1:
**Describe the bug**

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.
com.twelvemonkeys.imageio
imageio-webp
3.11.0

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 
Java 17.0.1 2021-10-19 LTS
Java(TM) SE Runtime Environment (build 17.0.1+12-LTS-39)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.1+12-LTS-39, mixed mode, sharing)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable.
JDK 17.0.1, IntelliJ Idea Community 2019.2.4, Win10 Home Basic

**To Reproduce**

1. Compile the below sample code
ImageIO.read(new URL("https://i0.hdslb.com/bfs/face/0266ce294d60018dcc6d567e57a3b93ae7825447.webp"));

2.Download the sample image file
https://i0.hdslb.com/bfs/face/0266ce294d60018dcc6d567e57a3b93ae7825447.webp

**Stak trace**
Always include the stack trace you experience.
java.lang.ArrayIndexOutOfBoundsException: Coordinate out of bounds!
	at java.desktop/sun.awt.image.ByteInterleavedRaster.setDataElements(ByteInterleavedRaster.java:541)
	at com.twelvemonkeys.imageio.plugins.webp.lossless.VP8LDecoder.decodeBwRef(VP8LDecoder.java:309)
	at com.twelvemonkeys.imageio.plugins.webp.lossless.VP8LDecoder.decodeImage(VP8LDecoder.java:225)
	at com.twelvemonkeys.imageio.plugins.webp.lossless.VP8LDecoder.readVP8Lossless(VP8LDecoder.java:131)
	at com.twelvemonkeys.imageio.plugins.webp.WebPImageReader.readVP8Lossless(WebPImageReader.java:673)
	at com.twelvemonkeys.imageio.plugins.webp.WebPImageReader.readAlpha(WebPImageReader.java:565)
	at com.twelvemonkeys.imageio.plugins.webp.WebPImageReader.readVP8Extended(WebPImageReader.java:498)
	at com.twelvemonkeys.imageio.plugins.webp.WebPImageReader.readVP8Extended(WebPImageReader.java:478)
[...]

Issue #739, opened 2023-03-16
WebP: Some images are throwing an ArrayIndexOutOfBoundsException when reading 


tc-wleite:
**Describe the bug**
I found a few WebP images that seem valid that are  are throwing an ArrayIndexOutOfBoundsException when reading.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use:
**3.9.4**. I also tested with the master branch.

2. The exact output of `java --version`: 
    openjdk version "11.0.13" 2021-10-19 LTS
    OpenJDK Runtime Environment (build 11.0.13+8-LTS)
    OpenJDK 64-Bit Server VM (build 11.0.13+8-LTS, mixed mode, sharing)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. 
I tested in Windows 11. 

**To Reproduce**
Run a simple standalone program (below), using the WebP ImageIO plugin.

**Expected behavior**
As the sample images I found seem to be valid (i.e. they are read without any trouble by ImageMagick, Google Chrome etc.), they should be read.

**Example code**
```java
import java.awt.image.BufferedImage;
import java.io.File;
import java.io.IOException;

import javax.imageio.ImageIO;

public class WebPReadTest {
    public static void main(String[] args) throws IOException {
        File input = new File("1-oob.webp"); // Input image
        File output = new File("1.png"); // Output image

        BufferedImage image = ImageIO.read(input);
        ImageIO.write(image, "png", output);
    }
}
```

**Sample files**
[sample-webp-images.zip](https://github.com/haraldk/TwelveMonkeys/files/10989612/sample-webp-images.zip)

**Stack trace**
Running with the test program using 3.9.4, I get the following exception:

```text
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Coordinate out of bounds!
	at java.desktop/sun.awt.image.ByteInterleavedRaster.setDataElements(ByteInterleavedRaster.java:541)
	at com.twelvemonkeys.imageio.plugins.webp.lossless.VP8LDecoder.decodeBwRef(VP8LDecoder.java:307)
	at com.twelvemonkeys.imageio.plugins.webp.lossless.VP8LDecoder.decodeImage(VP8LDecoder.java:223)
	at com.twelvemonkeys.imageio.plugins.webp.lossless.VP8LDecoder.readVP8Lossless(VP8LDecoder.java:131)
[...]

PR #1243, opened 2026-01-15
Fixes Lossless Alpha Channel WebP  

wezell:
Burned some midnight tokens but I believe this fixes the ongoing webp decoding issues with the VP8 java code, including:

https://github.com/haraldk/TwelveMonkeys/issues/978 and
https://github.com/haraldk/TwelveMonkeys/issues/739 

Much thanks to @Chris-SP365 for his [insight](https://github.com/haraldk/TwelveMonkeys/issues/739#issuecomment-2260573604) which allowed the LLM to generate a fix.  

### Fix
Aligned the lossless Huffman table construction with libwebp’s BuildHuffmanTable, which fixes the bitstream desync that was producing the shorter LZ77 lengths (e.g., 3584 vs 4095) and corrupting alpha. This switches the level‑2 table sizing to the same count‑based NextTableBitSize logic and keeps the single‑symbol 0‑bit behavior that libwebp uses, so symbol decoding and bit consumption now match the C decoder. I also added explicit bounds checks for backward references to throw IIOException instead of hitting AIOOB.

## Details:

Ported the table build to count‑driven sizing and replication so the Huffman table layout matches libwebp for tricky length distributions; this is the primary fix for lossless alpha corruption. HuffmanTable.java
Added guardrails around backward references to surface corrupt stream errors cleanly. VP8LDecoder.java


wezell:
This code contribution shall be licensed under the [BSD-3-Clause license](https://github.com/haraldk/TwelveMonkeys/tree/master#BSD-3-Clause-1-ov-file)

PR #1265, opened 2026-03-31
Avoid full-frame VP8 macroblock allocation for subsampled WebP reads

foryourever:
**What is fixed**

  Fixes excessive memory use when reading very large lossy WebP images with source subsampling, which could result in
  `OutOfMemoryError`.

  **Why is this change proposed**

  The VP8 decoder allocated macroblock state for the full frame up front, even when only a subsampled result was
  requested. For large images this caused unnecessary memory pressure and could make the read fail before the reduced
  output was produced.

  **What is changed**

  * Changed `VP8Frame` to decode with a bounded row-window working set instead of a full-frame macroblock allocation
  * Preserved the VP8 neighbor/context behavior needed for correct decoding
  * Added a repository-safe regression test for subsampled decode using the checked-in `alpha_filter.webp` sample

### DIFF
--- a/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/vp8/VP8Frame.java
+++ b/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/vp8/VP8Frame.java
@@ -164,9 +164,7 @@ public final class VP8Frame {
         currentMacroBlockRowIndex = -1;
         lastNonBPredMacroBlockByColumn = new MacroBlock[macroBlockCols + 2];
 
-        for (int mbCol = 0; mbCol < macroBlockCols + 2; mbCol++) {
-            lastNonBPredMacroBlockByColumn[mbCol] = topMacroBlockRow[mbCol];
-        }
+        System.arraycopy(topMacroBlockRow, 0, lastNonBPredMacroBlockByColumn, 0, macroBlockCols + 2);
     }
 
     private static int ySubBlockModeIndex(final int subX, final int subY) {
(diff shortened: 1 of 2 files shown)


foryourever:
收到

haraldk:
Thanks for the PR! Looks good, but I'll review in detail when I find some spare time. 

In the mean time, can you please look at the comments/suggestions from Sonar Cloud? At least 3 out of 4 should be quick fixes.

foryourever:
Thank you for the feedback! I've addressed the 3 SonarCloud suggestions and pushed the fixes. Please let me know if anything else needs adjustment. I really appreciate your time and look forward to your detailed review!





azhe
***@***.***

imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/vp8/VP8Frame.java
```diff
@@ -177,7 +262,14 @@ public boolean decode(final WritableRaster raster, final ImageReadParam param) t
 //        logger.log("macroBlockCols: " + macroBlockCols);
 //        logger.log("macroBlockRows: " + macroBlockRows);
 
-        createMacroBlocks();
+        windowedDecode = !debug;
```
haraldk, line 265: I'm not sure about the usefulness of this "debug mode" with a quite different code path... Do we need it? I'm afraid the the complexity of this flag outweighs the benefit...

haraldk (commented):
Sorry for taking way too long reviewing this. I just have a bit too much on my plat at the moment. 

I believe this to be a good addition to the reader! However, when reviewing this in-depth, I find the two separate modes of decoding quite frustrating. The complexity of the decoding is quite high already. I think it's better to remove the "non-windowed" (or debug) decoding. Unless you have very strong arguments why such a mode is needed..? 

I can also merge the changes as-is, and fix myself if you don't have the time to fix right now.

imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/vp8/VP8Frame.java
foryourever: You're absolutely right. The separate code path for debug mode introduces unnecessary complexity while offering little practical value. Feel free to delete it directly if it's convenient for you.

haraldk (approved):

PR #1275, opened 2026-05-09
[DDS] Adding writers for uncompressed formats

KhanhPham05:
Before I move to BC6H and BC7, which are a whole new realm of algorithm from traditional BC1-5, I want to take some times to add a writer for uncompressed format for DDS. All the currently supported `DDSType` now have a decoder and encoder.

# Code Changes
- field `DDSReader.RGB_16_ORDER` now called `DDSReader.RGB_565_ORDER` for 5:6:5 space to not causes confusion over 4:4:4:4 which also has 16 bits
- class `DDSType` : `blockSize` of uncompressed formats now represents the amount of _**bits**_ each pixel populates, not **_bytes_**
- class `DDSImageDataEncoder` : 
 + Now has 2 separate static methods to write compressed and uncompressed data
 + New nested class `UncompressedWriter` to write uncompressed colors, written as a reversed version of the corresponding uncompressed format decoder already existed in `DDSReader` class

# Notes
- "Consider swapping byte order to reflect the DXGI format?" - I think we should.
- As of BC6H and BC7, they are newer and less documented, some documentations describe it very vaguely to me so I need some times to research with a combination from LLMs and texted documentation to come to a functional decoder.
- Disclaimer: No AI code generation is used in the production of my code, the only A.I. usage is to research now encoders work in human text with researches from traditional texted documentations, the rest is hand-coded and troubleshot by hand.


KhanhPham05:
I think about `DDSImageWriteParam` we should have a way to allow the use of default uncompressed and compressed formats, which are A8R8G8B8 and BC3 ?!

haraldk:
Thank you for working on this and improving the writer! Much appreciated! 👍🏻 

Unfortunately, I don't have a lot of time to work on this at the moment, but I will review and hopefully merge your PR when I find some time. Just be patient. 😀 

In the mean time, please have a look at the failing tests and the new Sonar issues. 

About defaults, we can only have one default‡. And the most common default for ImageIO plugins is to write uncompressed, as that is usually lossless and fast (but not great for compression). We can also chose to use a compression as default, as long as its still lossless and fast. 

‡) In theory we can have one per image type, but it quickly becomes confusing...

imageio/imageio-dds/src/test/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageReaderTest.java
```diff
@@ -70,6 +70,7 @@ protected List<TestData> getTestData() {
         Dimension dim1 = new Dimension(1, 1);
 
         return Arrays.asList(
+                new TestData(getClassLoaderResource("/dds/dxt10.dds"), dim256),
```
haraldk: This file seems to be missing...

KhanhPham05:
I was looking at the `DDSReader` and I just want to figure out a way to change from using byte[] to just using the input **stream** (keyword: stream) because every time I see reading a whole byte[] into the memory, it freaks me out a bit.

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageWriter.java
```diff
@@ -25,7 +26,7 @@
 /**
  * ImageWriter implementation for Microsoft DirectDraw Surface (DDS) format.
  *
- * @author KhanTypo
+ * @author KhanhTypo
```
haraldk, line 28: ...a typo? 😆

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageWriter.java
```diff
@@ -116,6 +117,7 @@ else if (type != mipmapType) {
             processWarningOccurred(mipmapIndex, "All images in DDS mipmap must use same pixel format and compression");
         }
         if (mipmapType == null) {
+            //TODO: we can remove this one as we now have all formats write and read support.
```
haraldk: I agree, we should remove the exception.

But I think we need to change the logic slightly for uncompressed types, as `DDSImageWriteParam.type()` currently returns `null` for uncompressed types. Otherwise we don't properly enforce all mipmap sizes to use the same DDS type.

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageWriter.java
```diff
@@ -301,7 +304,7 @@ private void writePitchOrLinearSize(int height, int width, DDSType type) throws
             imageOutput.writeInt(((width + 3) / 4) * ((height + 3) / 4) * type.blockSize());
         }
         else {
-            imageOutput.writeInt(width * type.blockSize());
+            imageOutput.writeInt(Math.floorDiv(width * type.blockSize() + 7, 8));
```
haraldk: Why do you use `floorDiv` here instead of simple `/`?

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSReader.java
```diff
@@ -157,7 +157,7 @@ private int getBufferLength(DDSType type, int imageIndex) throws IIOException {
             case X8B8G8R8:
             case A8R8G8B8:
             case X8R8G8B8:
-                return type.blockSize() * width * height;
+                return (type.blockSize() / 8) * width * height;
```
haraldk: I don't think this is correct, unless the block size is an exact multiple of 8?

```suggestion
                return ((type.blockSize() + 7) / 8) * width * height;
```

i.e, the same logic as used for calculating the linear size in the Writer.

PS: I belive the block size is always a multiple of 8, so another option would be to just change back to using byte counts, and instead multiply the counts by 8 where bit count is needed.

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageWriteParam.java
```diff
@@ -12,9 +12,7 @@ public final class DDSImageWriteParam extends ImageWriteParam {
     private static final String[] COMPRESSION_TYPES = compressionTypes();
 
     private static String[] compressionTypes() {
-        // TODO: Maybe hardcode subset of values that we actually support writing?
         List<String> compressionTypes = Arrays.stream(DDSType.values())
-            .filter(DDSType::isBlockCompression)
```
haraldk, line 17: I'm a bit unsure about removing the filter here, as it re-introduces the confusion about DDS *encodings* and *compression*... 

The encodings are more related to the pixel layout. Not so much about compression... 

In ImageIO, it's often more "natural" to keep the pixel layout as close to the `BufferedImage` input as possible, to preserve pixel data. I.e. `TYPE_4BYTE_BGRA` or `TYPE_INT_ARGB` should encode as `A8R8G8B8`, `TYPE_INT_RGB` as `X8R8G8B8`, TYPE_USHORT_565_RGB` as `R5G6B5` etc, when compression is `None`.

haraldk (changes requested):
Thanks a lot for you work! 👍🏻 

I think we need to slightly rethink the concept of *compression* vs uncompressed *encodings*. I don't think we should mix the two. All the different types of uncompressed pixel layouts should use compression `None` and let the input `BufferedImage` decide the best encoding (to preserve pixel data as much as possible). This is also more in line with other ImageIO plugins.

Otherwise I think the changes are good!

haraldk:
And once agains, sorry for taking so long reviewing this...

haraldk:
Realizer I missed your comment about byte array vs stream when doing the review. But I completely agree. Reading the entire compressed image into a byte array buffer, before decoding is usually not the best way, as it is usually a waste of memory. Many short I/O reads on the other hand is usually a bad thing too (not a problem if we know the underlying stream is buffered, but er don't want several layers of buffering...). So, the best way is usually to read single rows of data, or perhaps single blocks for the block compressed formats.

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageWriter.java
KhanhPham05: it was my misunderstanding that if I make a div that has fraction that causes the value to be rounded up and make it off the expected value by one, turns out java discards the fraction if I cast straight from float to int so that was my misunderstanding., will fix :D

imageio/imageio-dds/src/main/java/com/twelvemonkeys/imageio/plugins/dds/DDSImageWriter.java
KhanhPham05: ironically, it is :D

KhanhPham05:
I have done a couple fix, I am testing some new logic on handling uncompressed encoding on my end.

KhanhPham05:
For now the proposed idea is to remove the ability for user to specify the uncompressed format via `DDSImageWriteParam.setCompressionType()` but to let user to set either `writeParam.setCompressionMode(ImageWriteParam.MODE_DISABLED)` or setting `compressionType = "None"`, the function `DDSType type()` will read the input image's color type and automatically return the corresponding `DDSType` to it.

The problem i am having now is that `ImageWriterBase.writeToSequence()` function uses `IIOImage` as input and it provides a function getting `RenderedImage` instead of `BufferedImage` so I can't get the image type normally but have to rely on `ColorModel` to decide, i don't want to add another layer of requirement by only allow input of `BufferedImage` but to support the `RenderedImage` itself.

KhanhPham05:
I tried this workaround and it seems to work fine.

Uncompressed formats are automatically recognised, only 8:8:8 color is allowed in the input image for now as I don't think people would usually use 5:6:5 or 5:5:5 and those make things trickier for me to work around the `ColorModel` and I don't think it worth spending too much time on them for now, I will try to add it in if we really need it

PR #1280, opened 2026-05-26
Fix out-of-bounds write in LZWDecoder.addStringToTable

metsw24-max:
Found this fuzzing the TIFF reader. A crafted LZW strip that keeps extending the dictionary without ever emitting a CLEAR code drives tableLength up to table.length, and the guard in addStringToTable lets the last entry through, so table[tableLength++] writes one slot past the array. A ~5 KB strip is enough to throw ArrayIndexOutOfBoundsException at index 4096 mid-decode. The added test fills the dictionary without clearing it.


haraldk (approved):

haraldk:
Thanks a lot for investigating and providing tests and fix! 👍🏻

PR #1281, opened 2026-05-27
validate ICC_PROFILE chunk index in getEmbeddedICCProfile

metsw24-max:
A JPEG with two or more APP2 'ICC_PROFILE' segments whose first chunk reports an index past the chunk count makes getEmbeddedICCProfile index the streams array out of range, throwing ArrayIndexOutOfBoundsException from read() before any pixels are decoded. The chunk number was only checked for being below 1, never against the count. Reject indexes outside 1..count and ignore the profile, matching the existing handling for other malformed ICC segments. The test builds such a file in memory.


haraldk:
Good catch! I'll review the PR when I find the time! 😀

metsw24-max:
Thanks for taking a look! Looking forward to your feedback.

haraldk (approved):

PR #1289, opened 2026-06-13
PR: Security hardening for XMPReader + SGI normalization fixes

mohd-gs:
Hi Harald,

I'd like to propose two small, targeted fixes — both non-breaking and minimal in scope:

**1. XMPReader: Disallow DOCTYPE declarations (XXE hardening)**

Added the `disallow-doctype-decl` feature as a defense-in-depth measure against XML External Entity (XXE) injection. Since XMP does not use DOCTYPE declarations in practice, this is a safe change and aligns with OWASP's recommended XML parser hardening guidelines.

**2. SGI Reader: Fix logic errors in `normalize()`**

Corrected several bugs causing silent pixel corruption:

- 16-bit path: fixed incorrect cast to `byte` (should be `short`) and wrong bitmask (`0xff` → `0xffff`).
- 8-bit path: changed `&&` to `||` in the threshold check, which was causing normalization to be skipped when only one bound was non-default.

These errors affected any 16-bit SGI image and 8-bit images with non-standard min/max values.

No API changes. Happy to add tests or clarify anything if needed.


haraldk (approved):
Perfect! Thanks

PR #1290, opened 2026-06-13
Re-enable disallow-doctype-decl in XMPReader (defense-in-depth)

mohd-gs:
This feature was disabled in f84b4589 because it broke `testNoExternalRequest` (which used XMP containing a DOCTYPE declaration).

This PR re-enables it and splits that test into two:

- **`testDoctypeDisallowed`** — verifies that XMP with a DOCTYPE is rejected.
- **`testNoExternalRequest`** — uses DOCTYPE-free XMP (with an `xml-stylesheet` PI referencing an external URL) to verify the remaining external-entity protections still hold.

XMP should never contain a DOCTYPE, so rejecting it outright is the safest defense-in-depth approach.


imageio/imageio-metadata/src/test/java/com/twelvemonkeys/imageio/metadata/xmp/XMPReaderTest.java
```diff
@@ -494,10 +494,24 @@ public void testRDFAttributeSyntaxCompoundDirectoryContentEXIF() throws IOExcept
...
     }
 
+    @Test
+    public void testDoctypeDisallowed() throws Exception {
+        // XMP with a DOCTYPE should be rejected now that disallow-doctype-decl is enabled
+        String xmlWithDoctype = resourceAsString("/xmp/xmp-jpeg-xxe.xml");
+
+        try (DirectImageInputStream input = new DirectImageInputStream(new ByteArrayInputStream(xmlWithDoctype.getBytes(StandardCharsets.UTF_8)))) {
+            createReader().read(input);
+            throw new AssertionError("XMP with DOCTYPE should have been rejected");
```
haraldk: Please use `fail(...)` from JUnit instead of throwing here. Or better yet, use `assertThrows` and verify the exception thrown.

imageio/imageio-metadata/src/test/java/com/twelvemonkeys/imageio/metadata/xmp/XMPReaderTest.java
```diff
@@ -494,10 +494,24 @@ public void testRDFAttributeSyntaxCompoundDirectoryContentEXIF() throws IOExcept
...
+        } catch (IOException expected) {
+            // Expected: parser rejects DOCTYPE declaration
```
haraldk: I think you need to assert on the exception, perhaps look at the cause type and verify that the message contains "DOCTYPE".

imageio/imageio-metadata/src/test/resources/xmp/xmp-jpeg-xxe-no-doctype.xml
```diff
@@ -0,0 +1,35 @@
+<?xpacket begin='﻿' id='W5M0MpCehiHzreSzNTczkc9d'?><?xml-stylesheet type="text/xsl" href="http://localhost:7777/xxx"?>
```
haraldk: Are you sure the test fails without disabling external entities now?

imageio/imageio-metadata/src/test/resources/xmp/xmp-jpeg-xxe-no-doctype.xml
haraldk, line 1: Maybe the file name should include "xml-stylesheet" instead of "no-doctype"?

haraldk (changes requested):

PR #1291, opened 2026-06-13
Fix operator precedence and normalization range in SGI normalize methods

mohd-gs:
Two bugs in the `normalize` methods of `SGIImageReader`:

1. **Operator precedence**: The `&` operator has lower precedence than `-` in Java, so expressions like `rowData[i] & 0xff - minValue` were evaluating as `rowData[i] & (0xff - minValue)` instead of `(rowData[i] & 0xff) - minValue`. This caused incorrect normalization whenever `PixMin` is non-zero.

2. **Normalization range**: The denominator was `maxValue` but should be `(maxValue - minValue)` to correctly map `[PixMin, PixMax]` → `[0, 0xff]` (or `[0, 0xffff]` for 16-bit). With `maxValue` as the denominator, `PixMax` would not map to full white when `PixMin` is non-zero.

Both bugs were invisible when `PixMin=0`. Added tests with non-zero `PixMin` values to cover these cases.


haraldk (approved):
Looks good!

PR #1292, opened 2026-06-22
bound RLE run length in ICNSUtil.decompress

metsw24-max:
**What is fixed**

An out-of-bounds write while decoding run-length-encoded ICNS icons. No open issue, found while reviewing the RLE path.

**Why is this change proposed**

`ICNSUtil.decompress` expands the RLE data of a compressed icon into a `result` array sized from the icon dimensions. Each run length comes from the compressed stream, but the bytes are written into `result` before anything checks the run still fits; the `remaining` counter is only decremented after the write. A crafted resource whose runs declare more bytes than the destination holds therefore overruns the array and an `ArrayIndexOutOfBoundsException` escapes straight out of `ImageReader.read()`.

The check belongs here rather than in the caller because the run length is only known once the run byte has been read inside this loop.

**What is changed**

* Reject a run whose length exceeds the remaining expected bytes with an `IIOException` before writing, for both the compressed and uncompressed branches, in line with the plugin's other malformed-input handling.
* Added `ICNSUtilTest` covering a valid run plus the two overflowing cases.

Should be a fairly contained fix; happy to adjust the message or exception type if you would prefer to clamp the run instead of rejecting it.

### DIFF
--- a/imageio/imageio-icns/src/main/java/com/twelvemonkeys/imageio/plugins/icns/ICNSUtil.java
+++ b/imageio/imageio-icns/src/main/java/com/twelvemonkeys/imageio/plugins/icns/ICNSUtil.java
@@ -30,6 +30,7 @@
 
 package com.twelvemonkeys.imageio.plugins.icns;
 
+import javax.imageio.IIOException;
 import java.io.DataInputStream;
 import java.io.IOException;
 
@@ -84,6 +85,10 @@ final class ICNSUtil {
                 // Compressed run
                 runLength = run + 131; // PackBits: -run + 1 and run == 0x80 is no-op... This allows 1 byte longer runs...
 
+                if (runLength > remaining) {
+                    throw new IIOException("Corrupt ICNS RLE data: run length exceeds expected data length");
+                }
+
                 byte runData = input.readByte();
 
                 for (int i = 0; i < runLength; i++) {
@@ -93,6 +98,10 @@ final class ICNSUtil {
             else {
                 // Uncompressed run
                 runLength = run + 1;
+
+                if (runLength > remaining) {
+                    throw new IIOException("Corrupt ICNS RLE data: run length exceeds expected data length");
+                }
 
                 input.readFully(result, resultPos, runLength);
                 resultPos += runLength;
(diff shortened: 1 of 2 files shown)


haraldk:
Thanks for the PR! The fix looks correct to me. 😀 

Have one question inline though.

imageio/imageio-icns/src/main/java/com/twelvemonkeys/imageio/plugins/icns/ICNSUtil.java
```diff
@@ -84,6 +85,10 @@ static void decompress(final DataInputStream input, final byte[] result, int off
                 // Compressed run
                 runLength = run + 131; // PackBits: -run + 1 and run == 0x80 is no-op... This allows 1 byte longer runs...
 
+                if (runLength > remaining) {
+                    throw new IIOException("Corrupt ICNS RLE data: run length exceeds expected data length");
+                }
```
haraldk, line 90: While this is strictly correct, I just wonder if just ignoring the overflow in the encoded runs would be more compatible? For compressed runs, it's simply 

    int len = Math.min(remaining, runLength);

For the compressed runs we would also have to discard the remaining bytes, ie.:

    input.skip(runLength - len);

Would that still be safe?

haraldk: I haven't seen any files with this issue in the wild though, maybe I'm just overthinking it... 😉

metsw24-max: Yeah, clamping is safe as far as the overflow goes: `len = Math.min(remaining, runLength)` keeps the write inside `result`, and the moment a run gets clamped `remaining` drops to 0, so the loop exits right after.

One detail though: only the uncompressed branch actually reads run-length bytes off the stream, so that's the only one that would need the `skip`. The compressed branch reads a single data byte and just repeats it, so clamping the copy count is enough there. And since the caller closes the wrapped stream as soon as decompress returns, the skip isn't strictly needed for alignment either (and `DataInputStream.skip` can short-read anyway).

My one hesitation is that clamping quietly produces a truncated channel, and once a run overruns the rest of the data is misaligned regardless, so it wouldn't really rescue a real file. Rejecting keeps it consistent with how the plugin handles other corrupt input. I'd lean towards keeping the reject, but happy to switch to the clamp + skip version if you'd rather err on the lenient side.

haraldk: Sorry about the confusion, I meant to `skip` only for the uncompressed runs, but somehow wrote compressed for both... 

Anyway, I think your argument is fair. Let's keep things clean and simple for now (and fix a later, if this turs about to be a real issue, it probably isn't...).

haraldk (approved):

PR #1304, opened 2026-07-15
clamp x delta to row bounds in BMP RLE8 decoder

arib06:
**What is fixed** The BMP RLE8 decoder writes out of bounds when a delta escape carries an x displacement larger than the row. `RLE8Decoder.decodeRow` computed `deltaX = srcX + stream.read()` from a raw stream byte and passed it as the `toIndex` of `Arrays.fill(row, srcX, deltaX, 0)` with no bound, so a crafted RLE8 bitmap (e.g. `00 02 FF 00` on a 4-byte row) throws `ArrayIndexOutOfBoundsException` out of the read path instead of failing as a normal `IIOException`.

**Why is this change proposed** The delta displacement crosses a trust boundary but is used unchecked, while every other opcode in the same method already guards its writes with `srcX < row.length`. The first displacement byte was also not passed through `checkEOF`, so a truncated stream produced a negative `toIndex` (`IllegalArgumentException`) rather than a clean EOF.

**What is changed**

* Clamp `deltaX` to `row.length` so the fill and the later reposition stay within the row
* Route the x displacement read through `checkEOF` so a truncated delta ends as `EOFException`
* Regression test in `RLE8DecoderTest` for an out-of-range delta

Valid RLE8 input is unaffected (a well-formed delta never moves x past the row), and the existing decoder tests still pass.

### DIFF
--- a/imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/RLE8Decoder.java
+++ b/imageio/imageio-bmp/src/main/java/com/twelvemonkeys/imageio/plugins/bmp/RLE8Decoder.java
@@ -75,7 +75,9 @@ final class RLE8Decoder extends AbstractRLEDecoder {
 
                     case 0x02:
                         // Delta
-                        deltaX = srcX + stream.read();
+                        // The x displacement comes straight from the stream; clamp it to the
+                        // row bounds so the fill and reposition below can't write past the row
+                        deltaX = Math.min(srcX + checkEOF(stream.read()), row.length);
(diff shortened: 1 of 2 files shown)


haraldk (approved):
I guess that makes sense, being lenient rather than throw exception, even if the input is malformed...

haraldk:
Thank you @arib06!

Appreciate your contributions and fixes! Sorry for taking so long, needed some time to catch up after a long summer break... 😀 

I'll try to make a release containing the fixes soon.

PR #1310, opened 2026-07-16
quote namespace prefix before regex match in SVGImageReaderSpi

arib06:
**What is fixed** `SVGImageReaderSpi.canDecodeInput` builds a regex from the tag name it reads off the input stream: `name.split(":")[0]` (up to 256 bytes of arbitrary input) is concatenated straight into the pattern passed to `String.matches`, and matched against a second attacker-controlled buffer. A prefix using nested bounded quantifiers, e.g. `<(a{0,100}){0,100}:svg xmlns:aaaa…=>`, compiles to a pattern with catastrophic backtracking, so a ~1 KB non-SVG input pins a CPU for many seconds inside the format sniff.

**Why is this change proposed** `canDecodeInput` runs for every registered provider on any stream handed to `ImageIO.read` / `getImageReaders`, before a reader is instantiated, so with imageio-batik on the classpath a single small upload wedges the calling thread. The prefix crosses a trust boundary but was used as a live pattern. `Pattern.quote` matches it as a literal, which also removes the `PatternSyntaxException` that an unbalanced prefix like `(` could throw out of `canDecodeInput`.

**What is changed**

* Wrap the namespace prefix in `Pattern.quote` before splicing it into the match pattern
* Add three crafted-prefix cases to `SVGImageReaderSpiTest` (they run under the existing 5s timeout guard, and hang before the fix)

Valid SVG detection is unchanged, and the existing batik tests still pass.

### DIFF
--- a/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReaderSpi.java
+++ b/imageio/imageio-batik/src/main/java/com/twelvemonkeys/imageio/plugins/svg/SVGImageReaderSpi.java
@@ -41,6 +41,7 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.util.Locale;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import static com.twelvemonkeys.imageio.util.IIOUtil.deregisterProvider;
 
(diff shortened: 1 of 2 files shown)


arib06:
any thoughts on this?

haraldk:
Looks good! 😀

haraldk (approved):

PR #1311, opened 2026-07-19
create temp cache files with owner-only permissions

arib06:
**What is fixed** No open issue; found while reading the stream caching code.

**Why is this change proposed** `File.createTempFile` opens with mode 0666 masked by the process umask, so with the common 022 umask the cache file lands as `rw-r--r--` in the shared temp directory. Both call sites write the full bytes of the stream being decoded, so on a multi-user host any local user can read the image data of everything that passes through `FileCacheSeekableStream`, or through the ICNS reader's `sips` JPEG 2000 conversion, for as long as the file exists. `Files.createTempFile` creates with `rw-------` instead, and `imageio-core`'s `FileCache` already uses it, so this just brings the two remaining helpers in line.

**What is changed**

* `FileCacheSeekableStream.createTempFile` and `SipsJP2Reader.dumpToFile` now create the temp file via `Files.createTempFile`
* Added two tests asserting owner-only permissions for the default and explicit-directory forms, skipped where POSIX attributes are unavailable
* Corrected the `FileCacheSeekableStream` class doc, which pointed at `File.createTempFile`

### DIFF
--- a/common/common-io/src/main/java/com/twelvemonkeys/io/FileCacheSeekableStream.java
+++ b/common/common-io/src/main/java/com/twelvemonkeys/io/FileCacheSeekableStream.java
@@ -33,17 +33,18 @@ package com.twelvemonkeys.io;
 import com.twelvemonkeys.lang.Validate;
 
 import java.io.*;
+import java.nio.file.Files;
 
 /**
  * A {@code SeekableInputStream} implementation that caches data in a temporary {@code File}.
  * <p>
- * Temporary files are created as specified in {@link File#createTempFile(String, String, java.io.File)}.
+ * Temporary files are created as specified in {@code Files.createTempFile}, and are readable by the owner only.
  * </p>
  *
  * @see MemoryCacheSeekableStream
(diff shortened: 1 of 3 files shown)


haraldk:
Thanks! The changes look reasonable, however Sonar complains, so I need to look into it in more detail to figure out how this change affects security.

arib06:
The Sonar flag is the "publicly writable directory" hotspot (S5443), which fires on any use of the default system temp dir. That directory choice is unchanged here: the old `File.createTempFile` resolved to the same `java.io.tmpdir`. The only behavioral difference is the file mode, 0644 (rw-r--r--) before and 0600 (rw-------) after, so it strictly narrows who can read the cached bytes.

It's a hotspot to review rather than a new vulnerability, and the same `Files.createTempFile` pattern is already used in imageio-core's FileCache. It can be marked reviewed/safe in Sonar to clear the gate. Happy to pass an explicit directory or add a suppression if you'd rather handle it in code.

haraldk (approved):

PR #1313, opened 2026-07-29
reject bit-plane counts over 8 in PCX planar row read

arib06:
**What is fixed** No open issue; hardening the PCX reader against a crafted header.

**Why is this change proposed** In the planar (bit-plane) branch of `PCXImageReader.read`, the row buffer is sized for exactly 8 planes (`planeWidth * 8`) and `BitRotator` always combines 8 planes, but the per-row fill length passed to `readRowByte` is `planeWidth * header.getChannels()`, where `channels` is an unsigned byte from the header. With `bitsPerPixel == 1` a channel count of 9-16 still builds a valid `IndexColorModel` (bit depth <= 16), so the branch runs and the fill length overruns the 8-plane buffer, leaking an `IndexOutOfBoundsException` out of `read()` instead of an `IOException` on a crafted file.

**What is changed**
* Reject bit-plane counts above 8 (the reader's fixed plane limit) with an `IIOException` before the planar row read; valid 2-8 plane PCX images decode unchanged.
* Added a regression test asserting `IIOException` for a 16-channel header.


haraldk (approved):

haraldk:
Thanks! Changes look good! I'll merge.

Issue #1102, opened 2025-03-04
WebP: Decoding makes the entire image look green


zly2006:
**Describe the bug**

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.

`com.twelvemonkeys.imageio:imageio-webp:3.12.0`

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 
For example: 
```
openjdk 23.0.2 2025-01-21
OpenJDK Runtime Environment Homebrew (build 23.0.2)
OpenJDK 64-Bit Server VM Homebrew (build 23.0.2, mixed mode, sharing)
```
**Example code**
```kotlin
    val z = ImageIO.read(File("sofa.webp"))
    // to png
    ImageIO.write(z, "png", File("sofa.png"))
```

**Sample file(s)**

[original webp & decoded png.zip](https://github.com/user-attachments/files/19067033/default.zip)

decoded png:
![Image](https://github.com/user-attachments/assets/ed81cedd-2538-46b7-a5b2-fff13e722016)

haraldk:
Looks like a duplicate of #734

haraldk:
Fixed via #1315

Issue #864, opened 2023-11-14
WebP: Images decoded with yellow tint


gotson:
**Describe the bug**
Some WebP images are decoded with a yellow tint.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.
3.10.1

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 

openjdk 21 2023-09-19 LTS
OpenJDK Runtime Environment Temurin-21+35 (build 21+35-LTS)
OpenJDK 64-Bit Server VM Temurin-21+35 (build 21+35-LTS, mixed mode)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable.

**To Reproduce**

1. Add the attached files as test resources for `imageio-webp`
2. Run the below unit tests
3. Check the generated PNG images, and compare to their JPG equivalent

**Expected behavior**
Images should look alike.

**Example code**

```java
@Test
    public void test1() throws IOException {
        testReadTestData(new TestData(getClassLoaderResource("/webp/Example1.webp"), new Dimension(2144, 3200)));
    }

    @Test
    public void test2() throws IOException {
        testReadTestData(new TestData(getClassLoaderResource("/webp/Example2.webp"), new Dimension(2144, 3200)));
    }

    public void testReadTestData(TestData data) throws IOException {
        ImageReader reader = createReader();

        reader.setInput(data.getInputStream());

        for (int i = 0; i < data.getImageCount(); i++) {
            BufferedImage image = null;

            try {
                image = reader.read(i);
                File tempActual = File.createTempFile("junit-testRead-", ".png");
                System.out.println("tempActual.getAbsolutePath(): " + tempActual.getAbsolutePath());
                ImageIO.write(image, "PNG", tempActual);
            }
            catch (Exception e) {
                e.printStackTrace();
                failBecause(String.format("Image %s index %s could not be read: %s", data.getInput(), i, e), e);
            }

            assertNotNull(String.format("Image %s index %s was null!", data.getInput(), i), image);

            assertEquals(
                String.format("Image %s index %s has wrong width: %s", data.getInput(), i, image.getWidth()),
                data.getDimension(i).width,
                image.getWidth()
            );
            assertEquals(
                String.format("Image %s index %s has wrong height: %s", data.getInput(), i, image.getHeight()),
                data.getDimension(i).height, image.getHeight()
            );
        }

        reader.dispose();
    }
```

[...]

Issue #785, opened 2023-07-08
WebP: Significant color distortion on "long" and lossy images


rzam:
**Describe the bug**
I'm noticing a color distortion when reading "big" (≥ ~4k-5k pixels in width or height) and lossy (generated with `cwebp -q 75`) WEBP images and converting them to other formats.

The distortion seems to be very significant when reading "long" (~5k+ pixels in height) WEBPs and less noticeable when reading "wide" (~5k+ pixels in width) WEBPs; WEBPs generated with a higher quality setting (eg. `cwebp -q 100`) seem unaffected.

**Version information**
TwelveMonkeys version (Gradle):
```
implementation("com.twelvemonkeys.imageio:imageio-webp:3.9.4")
```

Java version:
```
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode, sharing)
```

OS:
```
OS Name:                   Microsoft Windows 11 Pro
OS Version:                10.0.22000 N/A Build 22000
```

cwebp version:
```
1.3.1
libsharpyuv: 0.2.1
```

**Example code**
All images in the screenshots below were read and written using the standard `ImageIO` methods, eg.:
```
BufferedImage image = ImageIO.read(new File("long.cwebp.q75.webp"));
ImageIO.write(image, "png", new File("long.imageio.q75.png"));
```

**Sample file(s)**
[sources.zip](https://github.com/haraldk/TwelveMonkeys/files/11990442/sources.zip)
- `*.cwebp.*` files were generated with `cwebp`
- `*.imageio.*` were generated via the code above

**Screenshots**
Example using a "long" (500x5000) gradient:
![example long](https://github.com/haraldk/TwelveMonkeys/assets/6159707/0261db56-7aca-43e8-a612-b31a16e5a758)

Example using a "wide" (5000x500) gradient:
![example wide](https://github.com/haraldk/TwelveMonkeys/assets/6159707/91f5a66b-4276-4cd8-ac13-9bccbde6e7f9)

haraldk:
Hi, 

Thanks for reporting! There seems to be some kind of error that is accumulated in the vertical dimension, but I will need to investigate...

[...]

Issue #730, opened 2023-01-20
WebP: OOM for test file


BigMichi1:
**Describe the bug**
we have a sample webp file which causes an OOM and we want to understand if this is something we need to expect (reading webp files need a lot of memory) or if there is something not correct in the image reader.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.
3.9.4

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 
For example: 
```
    openjdk version "17.0.5" 2022-10-18
    OpenJDK Runtime Environment Temurin-17.0.5+8 (build 17.0.5+8)
    OpenJDK 64-Bit Server VM Temurin-17.0.5+8 (build 17.0.5+8, mixed mode, sharing)
```

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable.

**To Reproduce**

1. Compile the below sample code
2. Download the sample image file
3. Run the code with the sample file and Xmx set to 1024m or lower

**Expected behavior**
should work with Xmx lower than 1280m

**Example code**
```
    @Test
    public void triggerOOM() throws IOException {
        try (final var file = new ClassPathResource("sample-webp-file-for-testing.webp").getInputStream()) {
            final var image = ImageIO.read(file);
            assertThat(image).isNotNull();
        }
    }
```

Less is more. Don't add your entire project, only the code required to reproduce the problem. 😀 

**Sample file(s)**
[sample-webp-file-for-testing.zip](https://github.com/haraldk/TwelveMonkeys/files/10464503/sample-webp-file-for-testing.zip)

**Stak trace**

      java.lang.OutOfMemoryError: Java heap space
        at com.twelvemonkeys.imageio.plugins.webp.vp8.MacroBlock.predictUV(MacroBlock.java:492)
        at com.twelvemonkeys.imageio.plugins.webp.vp8.MacroBlock.dequantMacroBlock(MacroBlock.java:211)
        at com.twelvemonkeys.imageio.plugins.webp.vp8.VP8Frame.decodeMacroBlockRow(VP8Frame.java:366)
        at com.twelvemonkeys.imageio.plugins.webp.vp8.VP8Frame.decode(VP8Frame.java:346)
        at com.twelvemonkeys.imageio.plugins.webp.WebPImageReader.readVP8(WebPImageReader.java:657)
[...]

PR #1315, opened 2026-08-06
Fix Webp green tint vp8 intra prediction and tm pred

### LINKED ISSUE #734: WebP: Image decoded as green
**Describe the bug**
A specific webp image turns green when converted to jpeg or png.

**Version information**
1. The version of the TwelveMonkeys ImageIO library in use.

3.9.4. I also tried with 3.10.0-SNAPSHOT.

2. The *exact* output of `java --version` (or `java -version` for older Java releases). 

openjdk version "1.8.0_332"
OpenJDK Runtime Environment Corretto-8.332.08.1 (build 1.8.0_332-b08)
OpenJDK 64-Bit Server VM Corretto-8.332.08.1 (build 25.332-b08, mixed mode)

3. Extra information about OS version, server version, standalone program or web application packaging, executable wrapper, etc. Please state exact version numbers where applicable.

I tested on macOS 13.2.1.

User reported on Windows 11.

**Expected behavior**
The converted image should look like the original.

**Sample file(s)**

[001e.zip](https://github.com/haraldk/TwelveMonkeys/files/10894361/001e.zip)

**Screenshots**

**Additional context**
This was reported in https://github.com/gotson/komga/issues/1077

karstenw-ro:
Hi Harald,

I let Claude analyze and fix the green tint WebP issues and it managed to pin it down to two root causes described below:

**What is fixed.** Fixes #734, #1102, #864, #785. Also fixes the remaining colour error on the sample attached to #730 (the out-of-memory issue itself was already fixed on trunk before this PR). Two defects in TwelveMonkeys' lossy VP8 decoder (`imageio-webp`) that both produce a green/yellow tint on decode.

**What is changed.**
- *Fix 1:* Intra prediction read `dest`, the same buffer the loop filter had already modified in place. Since VP8 key frames are entirely intra-coded, every filtered pixel fed into prediction of macroblocks below/right, so the error compounded down the frame — clean at the top, increasingly green/yellow toward the bottom. Fixed by adding an unfiltered snapshot per sub-block (`SubBlock.recon`, captured in `setDest()` before the loop filter runs) and new accessors (`getRecon()`, `getReconPredict()`, `getMacroBlockReconPredict()`) so neighbour reads for prediction use the unfiltered copy, not `dest`.
- *Fix 2:* `TM_PRED` read an accessor returning an all-zero array for a non-existent (out-of-frame) neighbour, instead of the spec's border constants. At macroblock (0,0) this gives `0+0-0=0` instead of `129+127-127=129`, a constant -129 offset across the whole frame; chroma clamps to 0 and `(Y,0,0)` renders as pure green. Fixed by adding `getReconAbove()`, `getReconLeft()`, `getReconAboveLeft()`, returning unfiltered `recon` when the neighbour exists and 127/129 otherwise (with the above-left corner resolved per the spec's parenthetical). `TM_PRED`'s luma/chroma loops in `MacroBlock` switched to these. Builds on fix 1's `recon` machinery — the two ship as one PR.

**Where the correct behaviour was derived from.** RFC 6386 (the VP8 spec): section 5/15 establish that loop-filtered samples feed only inter prediction of later frames, never intra prediction within the same frame; section 12.2 specifies the 127 (above)/129 (left) border constants for out-of-bounds neighbours and that `TM_PRED`, unlike `DC_PRED`, must use them. Both fixes are corroborated by matching libwebp's reference implementation: it saves each row's bottom-edge samples before filtering for prediction, and pre-fills the borders with the same 127/129 constants and corner rule.

**How it was verified.** Each commit tested in isolation (clean worktrees off `41aaf3b1`) and combined, against 7 sample files from the related GitHub issues (compared to `dwebp -nofancy`) and the 131-file libwebp conformance corpus. Combined result: zero regressions across all 131 conformance files; mean error/channel on the issue samples drops from double digits (up to 114.17) to under 0.5. `mvn test -pl imageio/imageio-webp -am` on OpenJDK 26: 115 tests, 0 failures/errors, matching trunk on the same JDK — I have not yet run the full-reactor suite across the JDK 8/11/17/21/25 matrix this repo's CI uses, so that is still to be confirmed by CI itself.

**Which issue needs which commit.**
- #734 needs both commits: the green-tint sample needs commit 1; the fully-green `001e.webp` sample needs commit 2 as well
- #1102, #864, #785, #730 are fixed by commit 1 alone (mean error/channel: 5.75->0.41, 5.04->0.37, 21.30->0.18, 14.75->0.30 respectively)

### DIFF
--- a/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/vp8/MacroBlock.java
+++ b/imageio/imageio-webp/src/main/java/com/twelvemonkeys/imageio/plugins/webp/vp8/MacroBlock.java
@@ -577,11 +577,12 @@ final class MacroBlock {
             case Globals.TM_PRED:
                 // TODO:
                 // System.out.println("UV TM_PRED MB");
+                boolean aboveRowAvailableUV = this.y > 0;
                 MacroBlock ALMb = frame.getMacroBlock(x - 1, y - 1);
                 SubBlock ALUSb = ALMb.getUSubBlock(1, 1);
-                int alu = ALUSb.getRecon()[3][3];
+                int alu = ALUSb.getReconAboveLeft(aboveRowAvailableUV)[3][3];
                 SubBlock ALVSb = ALMb.getVSubBlock(1, 1);
-                int alv = ALVSb.getRecon()[3][3];
+                int alv = ALVSb.getReconAboveLeft(aboveRowAvailableUV)[3][3];
 
                 aboveUSb = new SubBlock[2];
                 leftUSb = new SubBlock[2];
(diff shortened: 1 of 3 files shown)


haraldk:
Thank you, this is great news! 🎉 

I will review the changes as soon as possible, but as I just got back from summer vacation, I have a bit of a backlog at the moment... 

In the mean time, please have a look at the Sonar issues (seems like easy-to-fix variable renaming).

karstenw-ro:
You're welcome!

I've fixed the names and the Sonar check passes now.

haraldk (approved):

PR #1317, opened 2026-08-19
add missing return after flat read in RGBE.readPixelsRawRLE

arib06:
**What is fixed** Out-of-bounds write in the HDR/RGBE scanline decoder when reading a crafted file.

**Why is this change proposed** `RGBE.readPixelsRawRLE` drops the two `return` statements the reference decoder has after its flat-read fallbacks, so once a scanline is read flat it falls through and decodes the same scanline again through the RLE path. A crafted HDR whose leading scanline quad marks it as non-RLE, with the width echoed in the last two bytes, drives the second pass to write past the `width * 4` row that `HDRImageReader` allocates, throwing `ArrayIndexOutOfBoundsException` at index `4 * width` from `ImageIO.read`. Widths below the RLE minimum reach the same fall-through.

**What is changed**

* Added the missing `return` after each flat read in `readPixelsRawRLE`, matching the reference implementation
* Added regression tests for the non-RLE and sub-minimum-width scanline cases

### DIFF
--- a/imageio/imageio-hdr/src/main/java/com/twelvemonkeys/imageio/plugins/hdr/RGBE.java
+++ b/imageio/imageio-hdr/src/main/java/com/twelvemonkeys/imageio/plugins/hdr/RGBE.java
@@ -219,6 +219,7 @@ final class RGBE {
         if ((scanline_width < 8) || (scanline_width > 0x7fff)) {
             // run length encoding is not allowed so read flat
             readPixelsRaw(in, data, offset, scanline_width * num_scanlines);
+            return;
         }
 
         // read in each successive scanline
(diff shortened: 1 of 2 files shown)


haraldk:
Thanks! 

Changes looks reasonable, I'll merge when I find some time. Looking at the code again, it's obvious that there should be returns there... 🤔 

Out of curiosity, where can I find the RGBE "reference decoder" you refer to? Is it the original C version?

arib06:
Yes, the original C version, Bruce Walter's rgbe.c (the one linked in this file's header comment, graphics.cornell.edu/~bjw/). In its RGBE_ReadPixels_RLE the two flat-read fallbacks, the sub-8/over-0x7fff width case and the non-RLE first-pixel case, both end with `return RGBE_ReadPixels(...)`. Those two returns are what got dropped in the Java port.

haraldk (approved):

PR #1323, opened 2026-08-24
validate item id and bound filename length in thumbs.db catalog

arib06:
**What is fixed** No open issue; hardening the Thumbs.db catalog reader. `Catalog.read` turns two values read straight from the OLE2 catalog stream into array positions without a bounds check, so a crafted Thumbs.db throws `ArrayIndexOutOfBoundsException` out of `ImageReader.read` instead of failing as an `IIOException`.

**Why is this change proposed** Both sites cross a trust boundary but use the stream value unchecked. In `CatalogItem.read` the item name is filled into a fixed 256 char buffer, so a name of 256 or more UTF-16 units before the terminator overruns it. In `Catalog.read` each item is stored at `items[item.getItemId() - 1]`, where `itemId` is a raw stream int, so an id of 0 (index -1) or one larger than the thumbnail count writes past the array. The surrounding reader already reports malformed input as `IIOException`.

**What is changed**

* Collect the filename into a `StringBuilder` instead of a fixed `char[256]`, so any length is read without overrun
* Validate the item id maps to a slot inside the catalog before the write, throwing `IIOException` when it does not
* Regression tests in `CatalogTest` for the long filename and out-of-range id, plus a valid catalog

Valid Thumbs.db files decode unchanged, and the existing reader tests still pass.

### DIFF
--- a/imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
+++ b/imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
@@ -34,6 +34,7 @@ import com.twelvemonkeys.io.LittleEndianDataInputStream;
 import com.twelvemonkeys.io.ole2.CompoundDocument;
 import com.twelvemonkeys.lang.StringUtil;
 
+import javax.imageio.IIOException;
 import java.io.DataInput;
 import java.io.IOException;
 import java.io.InputStream;
@@ -91,7 +92,11 @@ public final class Catalog implements Iterable<Catalog.CatalogItem> {
         for (int i = 0; i < header.getThumbnailCount(); i++) {
             CatalogItem item = CatalogItem.read(pDataInput);
             //System.out.println("item: " + item);
-            items[item.getItemId() - 1] = item;
+            int index = item.getItemId() - 1;
+            if (index < 0 || index >= items.length) {
+                throw new IIOException("Thumbs.db catalog item id out of range: " + item.getItemId());
+            }
+            items[index] = item;
         }
 
         return new Catalog(header, items);
@@ -225,15 +230,13 @@ public final class Catalog implements Iterable<Catalog.CatalogItem> {
 
             item.mLastModified = CompoundDocument.toJavaTimeInMillis(pDataInput.readLong());
 
-            char[] chars = new char[256];
+            StringBuilder name = new StringBuilder();
             char ch;
-            int len = 0;
             while ((ch = pDataInput.readChar()) != 0) {
-                chars[len++] = ch;
+                name.append(ch);
             }
 
-            String name = new String(chars, 0, len);
-            item.mFilename = StringUtil.getLastElement(name, "\\");
+            item.mFilename = StringUtil.getLastElement(name.toString(), "\\");
 
             item.mReserved2 = pDataInput.readShort();
             return item;
--- a/imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
+++ b/imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
@@ -230,7 +230,7 @@ public final class Catalog implements Iterable<Catalog.CatalogItem> {
 
             item.mLastModified = CompoundDocument.toJavaTimeInMillis(pDataInput.readLong());
 
-            StringBuilder name = new StringBuilder();
+            StringBuilder name = new StringBuilder(256);
             char ch;
             while ((ch = pDataInput.readChar()) != 0) {
                 name.append(ch);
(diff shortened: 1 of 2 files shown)


imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
```diff
@@ -225,15 +230,13 @@ private static CatalogItem read(final DataInput pDataInput) throws IOException {
 
             item.mLastModified = CompoundDocument.toJavaTimeInMillis(pDataInput.readLong());
 
-            char[] chars = new char[256];
+            StringBuilder name = new StringBuilder();
```
haraldk:
```suggestion
            StringBuilder name = new StringBuilder(256);
```

haraldk (dismissed):
Looks reasonable!

imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
```diff
@@ -225,15 +230,13 @@ private static CatalogItem read(final DataInput pDataInput) throws IOException {
 
             item.mLastModified = CompoundDocument.toJavaTimeInMillis(pDataInput.readLong());
 
-            char[] chars = new char[256];
+            StringBuilder name = new StringBuilder(256);
             char ch;
-            int len = 0;
             while ((ch = pDataInput.readChar()) != 0) {
```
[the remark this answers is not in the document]
haraldk: Or perhaps having a check inside the `while` loop, throwing an `IOException` if the name length exceeded  256 would be better.

imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
```diff
@@ -225,15 +230,13 @@ private static CatalogItem read(final DataInput pDataInput) throws IOException {
...
-                chars[len++] = ch;
+                name.append(ch);
             }
 
-            String name = new String(chars, 0, len);
-            item.mFilename = StringUtil.getLastElement(name, "\\");
+            item.mFilename = StringUtil.getLastElement(name.toString(), "\\");
 
             item.mReserved2 = pDataInput.readShort();
             return item;
```
haraldk, line 251: Did some more research on this, and this is probably reverse engineered and incorrect... 😛

The correct layout of the Catalog record is actually: 

```
offset  size       field
------  ---------  -------------------------------
0x00    4          record_size, little-endian
0x04    4          item_id, little-endian
0x08    8          timestamp, little-endian FILETIME-like value
0x10    N          UTF-16LE filename, NUL-terminated
```

So we should probably rewrite the entire parsing to use the record size, and read the correct number of chars directly (possibly skipping the null-terminator).

imageio/imageio-thumbsdb/src/test/java/com/twelvemonkeys/imageio/plugins/thumbsdb/CatalogTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private static void writeHeader(final LittleEndianDataOutputStream out, final int thumbCount) throws IOException {
+        out.writeShort(0); // reserved1
+        out.writeShort(0); // reserved2
+        out.writeInt(thumbCount);
+        out.writeInt(96); // max width
+        out.writeInt(96); // max height
+    }
+
+    private static void writeItem(final LittleEndianDataOutputStream out, final int itemId, final String name) throws IOException {
+        out.writeInt(0); // reserved1
```
haraldk: This should be item length in bytes.

imageio/imageio-thumbsdb/src/test/java/com/twelvemonkeys/imageio/plugins/thumbsdb/CatalogTest.java
```diff
@@ -0,0 +1,112 @@
...
+        out.writeInt(itemId);
+        out.writeLong(0L); // last modified
+        for (int i = 0; i < name.length(); i++) {
+            out.writeChar(name.charAt(i));
+        }
+        out.writeChar(0); // NUL terminator
+        out.writeShort(0); // reserved2
```
haraldk: Not sure why there is both NUL-terminator + a fixed 2 byte 0 here, but that seems to be the case with the sample files.

imageio/imageio-thumbsdb/src/main/java/com/twelvemonkeys/imageio/plugins/thumbsdb/Catalog.java
arib06: Good call. Reworked it to key off the record size instead of scanning to the NUL. The first int is now read as the record size, and the name comes from the remaining bytes (record size minus the 16 byte fixed part), stopping at the NUL and skipping the trailing padding. Records advance off the size rather than the terminator, so a missing NUL can't spill into the next record. The item id bounds check stays. All 91 tests still pass, including the two real sample files.

imageio/imageio-thumbsdb/src/test/java/com/twelvemonkeys/imageio/plugins/thumbsdb/CatalogTest.java
arib06: Fixed. The test now writes the real record size (16 byte fixed part + UTF-16 name + NUL + 2 byte padding) as the first field, matching what the reader reads.

haraldk (approved):

PR #1326, opened 2026-08-26
validate QuickTime RAW data size against pixel geometry

arib06:
**What is fixed** No open issue; hardening the PICT reader against a crafted CompressedQuickTime opcode.

**Why is this change proposed** In `QTRAWDecompressor.decompress` the pixel buffer is sized from the `idsc` Atom `dataSize` field, but the 32-bit branch then swaps bytes at `4 * y * width + x * 4`, where `width`/`height` come from separate 16-bit header fields. A file where `dataSize` is smaller than `width * height * 4` overruns the buffer (`ArrayIndexOutOfBoundsException` out of `ImageIO.read` rather than an `IIOException`); the 24- and 40-bit branches hit the same mismatch as a `RasterFormatException`.

**What is changed**

* Validate `width`/`height` are positive and that the declared data length covers `width * height * bytesPerPixel` before indexing, throwing `IIOException` on malformed input. The product is computed as `long` since the 16-bit dimensions can overflow `int` at depth 32.
* Added regression tests for the undersized-data case at depth 24, 32 and 40.

### DIFF
--- a/imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressor.java
+++ b/imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressor.java
@@ -67,6 +67,17 @@ final class QTRAWDecompressor extends QTDecompressor {
             dataStream.readFully(data, 0, description.dataSize);
         }
 
+        // Width, height and dataSize are independent fields from the 'idsc' Atom, so a crafted stream may
+        // declare a data size smaller than width * height * bytesPerPixel. Validate before indexing the buffer
+        // by pixel geometry below. Note width and height are 16 bit, so the product must be computed as long.
+        int bytesPerPixel = description.depth == 24 ? 3 : description.depth == 32 ? 4 : 1;
+        if (description.width <= 0 || description.height <= 0
+                || (long) description.width * description.height * bytesPerPixel > data.length) {
+            throw new IIOException(String.format(
+                    "Corrupt QuickTime RAW: data size %d too small for %dx%d at depth %d",
+                    data.length, description.width, description.height, description.depth));
+        }
+
         DataBuffer buffer = new DataBufferByte(data, data.length);
 
         WritableRaster raster;
--- a/imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressor.java
+++ b/imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressor.java
@@ -61,21 +61,22 @@ final class QTRAWDecompressor extends QTDecompressor {
     }
 
     public BufferedImage decompress(final ImageDesc description, final InputStream stream) throws IOException {
+        // Width, height and dataSize are independent fields from the 'idsc' Atom, so a crafted stream may
+        // declare a data size smaller than width * height * bytesPerPixel. Validate before allocating and
+        // indexing the buffer by pixel geometry below. Note width and height are unsigned 16 bit, so the
+        // product must be computed as long.
+        int bytesPerPixel = description.depth == 24 ? 3 : description.depth == 32 ? 4 : 1;
+        if (description.width <= 0 || description.height <= 0
+                || (long) description.width * description.height * bytesPerPixel > description.dataSize) {
+            throw new IIOException(String.format(
+                    "Corrupt QuickTime RAW: data size %d too small for %dx%d at depth %d",
+                    description.dataSize, description.width, description.height, description.depth));
+        }
+
         byte[] data = new byte[description.dataSize];
 
         try (DataInputStream dataStream = new DataInputStream(stream)) {
             dataStream.readFully(data, 0, description.dataSize);
-        }
-
-        // Width, height and dataSize are independent fields from the 'idsc' Atom, so a crafted stream may
-        // declare a data size smaller than width * height * bytesPerPixel. Validate before indexing the buffer
-        // by pixel geometry below. Note width and height are 16 bit, so the product must be computed as long.
-        int bytesPerPixel = description.depth == 24 ? 3 : description.depth == 32 ? 4 : 1;
-        if (description.width <= 0 || description.height <= 0
-                || (long) description.width * description.height * bytesPerPixel > data.length) {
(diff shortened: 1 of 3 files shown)


imageio/imageio-pict/src/main/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressor.java
```diff
@@ -67,6 +67,17 @@ public BufferedImage decompress(final ImageDesc description, final InputStream s
             dataStream.readFully(data, 0, description.dataSize);
         }
 
+        // Width, height and dataSize are independent fields from the 'idsc' Atom, so a crafted stream may
+        // declare a data size smaller than width * height * bytesPerPixel. Validate before indexing the buffer
+        // by pixel geometry below. Note width and height are 16 bit, so the product must be computed as long.
+        int bytesPerPixel = description.depth == 24 ? 3 : description.depth == 32 ? 4 : 1;
+        if (description.width <= 0 || description.height <= 0
```
haraldk, line 69: I believe these values should have been read as unsigned short values, making negative values impossible. Otherwise, validating the size sounds like a good idea! 👍🏻 

Unfortunately, I don't find a formal description of the format of the `idsc`atom, Apple's own documentation only contains an example... So I can't say for sure if values > 32767 are allowed as image dimensions.

haraldk: Another thing, as `data.length` is always `description.dataSize` we can move the validation before even allocating the array, making things even safer.

arib06: Both good points. Switched the width/height reads in ImageDesc to readUnsignedShort, so negatives can't happen and the full 16-bit range is allowed since the format isn't nailed down. And moved the size check ahead of the allocation, validating against description.dataSize directly. The <= 0 guard now just catches zero dimensions.

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
haraldk:
```suggestion
class QTRAWDecompressorTest {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
```diff
@@ -42,4 +46,41 @@ public void canDecompressGray() {
 
         assertTrue(decompressor.canDecompress(createDescription(40)));
     }
+
+    @Test
+    public void decompressRGBADataSizeTooSmall() {
```
haraldk:
```suggestion
    void decompressRGBADataSizeTooSmall() {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
```diff
@@ -42,4 +46,41 @@ public void canDecompressGray() {
...
+        description.height = 8;
+        description.dataSize = 16;
+
+        QTDecompressor decompressor = new QTRAWDecompressor();
+        assertThrows(IIOException.class,
+                () -> decompressor.decompress(description, new ByteArrayInputStream(new byte[description.dataSize])));
+    }
+
+    @Test
+    public void decompressRGBDataSizeTooSmall() {
```
haraldk:
```suggestion
    void decompressRGBDataSizeTooSmall() {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
```diff
@@ -42,4 +46,41 @@ public void canDecompressGray() {
...
+        description.height = 8;
+        description.dataSize = 16;
+
+        QTDecompressor decompressor = new QTRAWDecompressor();
+        assertThrows(IIOException.class,
+                () -> decompressor.decompress(description, new ByteArrayInputStream(new byte[description.dataSize])));
+    }
+
+    @Test
+    public void decompressGrayDataSizeTooSmall() {
```
haraldk:
```suggestion
    void decompressGrayDataSizeTooSmall() {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
haraldk:
```suggestion
class QTRAWDecompressorTest {
```

haraldk:
```suggestion
    void canDecompressRGB() {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
haraldk:
```suggestion
    void decompressRGBADataSizeTooSmall() {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
haraldk:
```suggestion
    void decompressRGBDataSizeTooSmall() {
```

imageio/imageio-pict/src/test/java/com/twelvemonkeys/imageio/plugins/pict/QTRAWDecompressorTest.java
haraldk:
```suggestion
    void decompressGrayDataSizeTooSmall() {
```

haraldk (approved):

PR #1327, opened 2026-08-27
reject out-of-range palette register in iff multipalette color map

arib06:
**What is fixed** No open issue; hardening the IFF multipalette reader against a crafted PCHG chunk.

**Why is this change proposed** `MutableIndexColorModel.adjustColorMap` validates palette change registers only against the upper bound (`index >= rgbs.length`) before writing `rgbs[index]`, so a negative register slips through and writes out of bounds. ILBM `PCHG` BigLineChanges read a 32-bit register straight from the stream with a signed `toShort`, so a register like `0xfffe` becomes `-2` and reaches `adjustColorMap` through `getColorModel`/`colorMapForRow` while rendering, throwing `ArrayIndexOutOfBoundsException` out of `ImageIO.read` on a crafted IFF.

**What is changed**

* Reject registers below zero as well as above `rgbs.length` at the existing validation point, matching the current warn-and-ignore behaviour for out-of-range registers.
* Keep the `MP_REG_IGNORE` sentinel silently skipped so valid multipalette decoding is unchanged.
* Added `PCHGChunkTest` covering a negative register (ignored) and an in-range register (applied).

### DIFF
--- a/imageio/imageio-iff/src/test/java/com/twelvemonkeys/imageio/plugins/iff/PCHGChunkTest.java
+++ b/imageio/imageio-iff/src/test/java/com/twelvemonkeys/imageio/plugins/iff/PCHGChunkTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, Harald Kuhr
+ * Copyright (c) 2026, Harald Kuhr
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -43,7 +43,7 @@ import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-public class PCHGChunkTest {
+class PCHGChunkTest {
 
     private static IndexColorModel base() {
         return new IndexColorModel(4, 16, new int[16], 0, false, -1, DataBuffer.TYPE_BYTE);
@@ -77,7 +77,7 @@ public class PCHGChunkTest {
     }
 
     @Test
-    public void negativeRegisterIsIgnored() throws IOException {
+    void negativeRegisterIsIgnored() throws IOException {
         // reg 0xfffe reads back as a negative short; adjustColorMap must not index the palette with it
         PCHGChunk chunk = parse(chunkWithRegister(0xfffe));
         IndexColorModel base = base();
@@ -88,7 +88,7 @@ public class PCHGChunkTest {
     }
 
     @Test
-    public void inRangeRegisterIsApplied() throws IOException {
+    void inRangeRegisterIsApplied() throws IOException {
         PCHGChunk chunk = parse(chunkWithRegister(5));
         IndexColorModel base = base();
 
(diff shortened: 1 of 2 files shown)


imageio/imageio-iff/src/test/java/com/twelvemonkeys/imageio/plugins/iff/PCHGChunkTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.awt.image.IndexColorModel;
+import java.io.ByteArrayInputStream;
+import java.io.DataInputStream;
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class PCHGChunkTest {
```
haraldk:
```suggestion
class PCHGChunkTest {
```

imageio/imageio-iff/src/test/java/com/twelvemonkeys/imageio/plugins/iff/PCHGChunkTest.java
```diff
@@ -0,0 +1,100 @@
...
+                0x00, 0x00, 0x00, 0x01, // totalChanges = 1
+                0x40, 0x00, 0x00, 0x00, // line mask: only row 1 changed
+                0x00, 0x01,             // changeCount = 1
+                (byte) ((reg >> 8) & 0xff), (byte) (reg & 0xff), // register
+                0x00, 0x11, 0x22, 0x33, // alpha (skipped), r, b, g
+        };
+    }
+
+    @Test
+    public void negativeRegisterIsIgnored() throws IOException {
```
haraldk:
```suggestion
    void negativeRegisterIsIgnored() throws IOException {
```

imageio/imageio-iff/src/test/java/com/twelvemonkeys/imageio/plugins/iff/PCHGChunkTest.java
```diff
@@ -0,0 +1,100 @@
...
+        PCHGChunk chunk = parse(chunkWithRegister(0xfffe));
+        IndexColorModel base = base();
+
+        ColorModel palette = assertDoesNotThrow(() -> chunk.getColorModel(base, 1, false));
+
+        assertNotNull(palette);
+    }
+
+    @Test
+    public void inRangeRegisterIsApplied() throws IOException {
```
haraldk:
```suggestion
    void inRangeRegisterIsApplied() throws IOException {
```

imageio/imageio-iff/src/test/java/com/twelvemonkeys/imageio/plugins/iff/PCHGChunkTest.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright (c) 2024, Harald Kuhr
```
haraldk:
```suggestion
 * Copyright (c) 2026, Harald Kuhr
```

😉

haraldk (dismissed):
Thanks! Appreciate your fixes! 

In this case it does fix the very specific issue of negative indexes. 👍🏻 

However, I think it would provide a lot more value if we could also fix the issue as in the comment:

> // TODO: Move validation to chunk (when reading)

Would you like to look into that as well?

haraldk (approved):

PR #1333, opened 2026-08-31
Fix #1019: don't add alpha channel in createRotated for quadrant rotations

### LINKED ISSUE #1019: ImageUtil.createRotated() unexpectedly adds an alpha channel
Hello,

TwelveMonkeys version: `3.10.1`

`ImageUtil.createRotated()` unexpectedly adds an alpha channel to the result image
when the source image does not have an alpha channel.

For example:

```java
BufferedImage source = new BufferedImage(100, 50, BufferedImage.TYPE_INT_RGB);
assertEquals(100, source.getWidth());
assertEquals(50, source.getHeight());
assertFalse(source.getColorModel().hasAlpha());

BufferedImage result = ImageUtil.createRotated(source, 90);
assertEquals(50, result.getWidth());
assertEquals(100, result.getHeight());
assertFalse(result.getColorModel().hasAlpha()); // Unexpectedly fails
```

I noticed this after rotating a JPEG source image
then saving the result image as a JPEG
produces a `IIOException: Bogus input colorspace`

Obviously we can manually remove the alpha channel
after using `ImageUtil.createRotated()`

But it is suprising that an alpha channel has been added to the result image
after rotating a source image which does not have an alpha channel.

Thanks.

nandhak0318:
A quadrant rotation (90/180/270 degrees) covers every destination pixel with source pixels, so the destination now keeps the source's color model instead of always being created as TYPE_INT_ARGB. Non-quadrant rotations still get a transparent destination, as the uncovered corners need alpha.

### DIFF
--- a/common/common-image/src/main/java/com/twelvemonkeys/image/ImageUtil.java
+++ b/common/common-image/src/main/java/com/twelvemonkeys/image/ImageUtil.java
@@ -721,8 +721,17 @@ public final class ImageUtil {
         AffineTransform transform = AffineTransform.getTranslateInstance((newW - w) / 2.0, (newH - h) / 2.0);
         transform.rotate(pAngle, w / 2.0, h / 2.0);
 
-        // TODO: Figure out if this is correct
-        BufferedImage dest = createTransparent(newW, newH);
+        // A quadrant rotation is completely covered by source pixels,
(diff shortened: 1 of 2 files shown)


haraldk (approved):
Looks reasonable! 👍🏻

PR #1341, opened 2026-09-08
Openize heic plugin updates

pborissow:
Synced with main, fixed destination offset handling, and added thumbnail and metadata support.


haraldk:
Thanks! I'll review this PR at some time, and will decide whether we want to include a third-party based plugin (which kind of breaks with the project's goals of delivering standalone plugins). 

PS: I looked at the last commit, and I think the addition of GPS and more Exif metadata can be useful, but perhaps better as a separate PR as it's not directly HEIC related? Also, I'm not sure if adding all the fields to the standard metadata "just in case" is a good idea. Most plugins sticks to a subset, I think the one from the previous commit (7c19ac1978935e96a0398f4a9e54f06f9bb07159) looks reasonable. 😀

